PR #28, opened 2025-03-20
supabase template update [WIP]

djin-lucas:
- Removed line with version: version: "3.8".
- Add container prefix with 7 char length hash.


blueprints/supabase/docker-compose.yml
```diff
@@ -5,11 +5,10 @@
 #   Destroy:        docker compose -f docker-compose.yml -f ../files/dev/docker-compose.dev.yml down -v --remove-orphans
 
 name: supabase
-version: "3.8"
 
 services:
   studio:
-    container_name: supabase-studio
+    container_name: ${CONTAINER_PREFIX}-supabase-studio
```
Siumauricio, line 11: Why do you need to add a prefix?

Siumauricio (changes requested):

blueprints/supabase/docker-compose.yml
djin-lucas: I was trying to create two supabase services in my project, but it was giving me conflict with other existing containers, is this expected? or did I do something wrong?

Issue #34, opened 2025-03-23
Add template for Backrest


maelp:
### What problem will this feature address?

Basic template for backup service Backrest to add to the app templates

```
version: "3.8"

services:
  backrest:
    image: garethgeorge/backrest:v1.7.3
    container_name: backrest
    hostname: backrest
    restart: unless-stopped
    ports:
      - "9898:9898"
    environment:
      - BACKREST_PORT=9898
      - BACKREST_DATA=/data
      - BACKREST_CONFIG=/config/config.json
      - XDG_CACHE_HOME=/cache
      - TZ=America/Los_Angeles
    volumes:
      - /mnt/data/backrest/data:/data
      - /mnt/data/backrest/config:/config
      - /opt/backrest/cache:/cache
      - /mnt/data:/userdata:ro # we mount /mnt/data (or whatever folder you want) to /userdata (that we want to backup)
```

### Describe the solution you'd like

Add a template

### Describe alternatives you've considered

Do it manually

### Additional context

Could be interesting to have default backups for Docker volumes and/or a `/mnt/data` disk for dokploy (this is where I store all the data for my containers, I find it "safer" than docker volumes which are a bit less easy to look at and manage)

### Will you send a PR to implement it?

Maybe, need help

PR #36, opened 2025-03-24
Add Backrest web backup

maelp:
This adds a template for Backrest


blueprints/backrest/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.8"
+services:
+  backrest:
+    image: garethgeorge/backrest:v1.7.3
+    restart: unless-stopped
+    ports:
+      - "9898:9898"
```
drudge:
```suggestion
      - 9898
```

blueprints/backrest/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.8"
```
drudge:
```suggestion
```

maelp:
Fixes https://github.com/Dokploy/templates/issues/34

blueprints/backrest/docker-compose.yml
```diff
@@ -11,11 +11,11 @@ services:
       - XDG_CACHE_HOME=/cache # backup cache
       - TZ=${TZ}
     volumes:
-      - backrest/data:/data
-      - backrest/config:/config
+      - backrest:/data
+      - backrest:/config
```
maelp: are you sure you want to mount the /data folder and the /config folder at the same location? will that work? there might be overwritten files no? @Siumauricio

meta.json
```diff
@@ -1469,3 +1469,1848 @@
     "tags": ["backup"]
   }
 ]
+=======
```
maelp: I think there are still git markers in the file @Siumauricio

blueprints/backrest/docker-compose.yml
Siumauricio: It seems that our internal parser of isolated deployment, can't parse correctly the volume in that way, it generates it in a wrong way, I think we'd better wait for it to fix the error.

meta.json
Siumauricio: Fixed

blueprints/backrest/docker-compose.yml
maelp: ah weird! Otherwise you can try creating a backrest-data and a backrest-config volume, it's the same

Siumauricio: Yes, that could be another quicker solution, but I would like to eliminate that case.

Siumauricio (approved):

blueprints/backrest/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    environment:
+      - BACKREST_PORT=9898
+      - BACKREST_DATA=/data # where it stores current state data
+      - BACKREST_CONFIG=/config/config.json # where it stores backup configurations
+      - XDG_CACHE_HOME=/cache # backup cache
+      - TZ=${TZ}
+    volumes:
+      - backrest/data:/data
+      - backrest/config:/config
+      - backrest/config:/cache # to preserve backup cache between restarts
```
maelp, line 16: I guess you should perhaps use `backrest/cache` here, or even the `backrest-cache` volume (even better)

Issue #37, opened 2025-03-24
Add template for ownCloud infinitescale


maelp:
```yaml
version: "3.7"

# Check https://fariszr.com/owncloud-infinite-scale-docker-setup/
services:
  ocis:
    image: owncloud/ocis:7.1.1
    user: 0:0
    ports:
      - 127.0.0.1:9200:9200
    entrypoint:
      - /bin/sh
    # run ocis init to initialize a configuration file with random secrets
    # it will fail on subsequent runs, because the config file already exists
    # therefore we ignore the error and then start the ocis server
    command: ["-c", "ocis init || true; ocis server"]
    environment:
      OCIS_URL: "https://ocis.xn--l3haa.st"
      OCIS_LOG_LEVEL: info
      PROXY_TLS: false
      OCIS_INSECURE: true
      IDM_ADMIN_PASSWORD: ${IDM_ADMIN_PASSWORD}
      SETTINGS_GRPC_ADDR: 0.0.0.0:9191
      GATEWAY_GRPC_ADDR: 0.0.0.0:9142 # make the REVA gateway accessible to the app drivers
      # PROXY_TLS is set to "false", the download url has no https
      STORAGE_USERS_DATA_GATEWAY_URL: http://ocis:9200/data
      THUMBNAILS_FILESYSTEMSTORAGE_ROOT: /var/lib/ocis-thumbnails
    volumes:
      - owncloud-ocis/ocis.yaml:/etc/ocis/ocis.yaml
      - owncloud-ocis/data:/var/lib/ocis
      - owncloud-ocis/thumbnails:/var/lib/ocis-thumbnails
```

and

```
IDM_ADMIN_PASSWORD: somePaswordForAdmin
```

BinkyTwin:
Opened PR #888 for the ownCloud Infinite Scale template.

/attempt #37

PR #38, opened 2025-03-24
Add ownCloud InfiniteScale

maelp:
Add ownCloud InfiniteScale (file-sharing)


maelp:
Fixes https://github.com/Dokploy/templates/issues/37

maelp:
PS: not sure if that's meant, but in the meta.json we fill the template "website", but in the view there is only "github" and "docs"?

Siumauricio (changes requested):
Getting this error 
Screenshot 2025-03-30 at 3 43 42 AM

maelp:
Hmmm weird, the "ocis init" in the initial docker-compose template should create that file initially...
CleanShot 2025-03-30 at 11 47 37@2x

maelp:
@Siumauricio I guess it's because the docker linter thinks that "owncloud/ocis.yaml" is the name of the volume, rather than "the file ocis.yaml inside the volume owncloud"

Siumauricio:
Yes @maelp , but when running docker compose, it won't know that it really exists, the file must be created previously, I think...

maelp:
@Siumauricio hmmm... would be nice then perhaps to add a way to add prepopulated file in the template? is that possible? Or just an empty file?

Siumauricio:
Yeah @maelp  you can create a empty file, using the [[config.mounts]] ..., just like this https://github.com/Dokploy/templates/blob/main/blueprints/plausible/template.toml#L16

maelp:
@Siumauricio okay I added the file

Siumauricio:
Same error
Screenshot 2025-04-12 at 1 53 24 AM

Siumauricio:
Closed due to inactivity.

PR #39, opened 2025-03-25
Add Kutt link shortener (much better than Shlink)

maelp:
This adds a modern https://kutt.it link shortener


Siumauricio (changes requested):
I'm getting this error
image

maelp:
Hmmm weird I guess it's related to how we create volumes? on my own deploy, I did it with a local disk mount rather than a Docker volume and it works, I don't know if this is possible with Docker volulmes

CleanShot 2025-03-30 at 11 48 43@2x

maelp:
Like if we create a "kutt" volume as done in the docker-compose, can we just set "kutt/db_data_sqlite:/var/lib/kutt" as a mountpoint in the docker-compose, or that's not allowed, and we should create a separate volulme "kutt-db-data" volume for the db data, and a kutt-custom volume for the custom data?

rubixvi:
```
services:
  kutt:
    image: kutt/kutt:v3.2.3
    volumes:
      - kutt_db_data:/var/lib/kutt
      - ./kutt/custom:/kutt/custom
    environment:
      DB_FILENAME: "/var/lib/kutt/data.sqlite"
      JWT_SECRET: ${JWT_SECRET}
      DEFAULT_DOMAIN: ${DOMAIN}
      CUSTOM_DOMAIN_USE_HTTPS: true

volumes:
  kutt_db_data: {}
  ```
  Try this

Siumauricio:
Is this PR still alive?

the problem is still there, besides it needs to be migrated from .yml to .toml

PR #40, opened 2025-03-26
Update to Latest supabase 1.25.02#7709e8d

malko:
This will update supabase template to latest version (ie: 1.25.02#7709e8d), this will notably add supavisor.

### DIFF
--- a/blueprints/supabase/docker-compose.yml
+++ b/blueprints/supabase/docker-compose.yml
@@ -1,9 +1,14 @@
-# Usage
-#   Start:              docker compose up
-#   With helpers:       docker compose -f docker-compose.yml -f ../files/dev/docker-compose.dev.yml up
-#   Stop:               docker compose down
-#   Destroy:            docker compose -f docker-compose.yml -f ../files/dev/docker-compose.dev.yml down -v --remove-orphans
-#   Reset everything:  ./reset.sh
+# Note: env passwords and keys are randomly generated at template install time
+#       YOU MUST REPLACE THE FOLLOWING VARIABLES:
+#         - JWT_SECRET
+#         - ANON_KEY
+#         - SERVICE_ROLE_KEY
+# Go to https://supabase.com/docs/guides/self-hosting/docker#securing-your-services
+# for more information on how to generate these keys
+#
+# API_EXTERNAL_URL and SUPABASE_PUBLIC_URL should be changed if you are using a custom domain
+#
+# If you are using an external Postgres database, you can disable the db service and update the DATABASE_URL in the supavisor service
 
 name: supabase
 
--- a/blueprints/supabase/docker-compose.yml
+++ b/blueprints/supabase/docker-compose.yml
@@ -1,17 +1,16 @@
 # Usage
-#   Start:          docker compose up
-#   With helpers:   docker compose -f docker-compose.yml -f ../files/dev/docker-compose.dev.yml up
-#   Stop:           docker compose down
-#   Destroy:        docker compose -f docker-compose.yml -f ../files/dev/docker-compose.dev.yml down -v --remove-orphans
+#   Start:              docker compose up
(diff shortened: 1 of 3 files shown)


blueprints/supabase/docker-compose.yml
```diff
@@ -428,21 +449,72 @@ services:
...
+      ]
+    security_opt:
+      - "label=disable"
+
+  # Update the DATABASE_URL if you are using an external Postgres database
+  supavisor:
+    image: supabase/supavisor:2.4.14
+    restart: unless-stopped
+    ports:
+      - ${POSTGRES_PORT}:5432
```
djin-lucas, line 474: This is causing this error, commenting this works correctly:

Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint meu-projeto-supabase-hs6i9o-supavisor-1 (): Bind for 0.0.0.0:5432 failed: port is already allocated

malko: Weird worked on my side will try to reproduce tonight

malko: @djin-lucas Did you change some settings from the default config ? I've just tried again a fresh deploy on a new project without this issue, the only settings i've changed from the default for first run was API_EXTERNAL_URL and SUPABASE_PUBLIC_URL

Accessing from another host using the connection strings as described here works for session based and pooled connections.

Is it possible that the problem may be related to you having another service running on that port ? If this is the case can you try to change the env value of 
POSTGRES_PORT=5432
POOLER_PROXY_PORT_TRANSACTION=6543

malko:
I passed it in draft as for now we still faces an issue with gotrue that won't send mails this is related to GOTRUE_MAILER_EXTERNAL_HOSTS apparently but even adding this env var doesn't seems to solve the problem

malko:
posted an issue in the gotrue repository about that https://github.com/supabase/auth/issues/1970

maelp:
I don't really understand why ANON_KEY and SERVICE_KEY are hardcoded in the config? shouldn't they be JWT tokens signed with the (random) JWT_SECRET?

maelp:
It seems to me we should embed a script to run to generate the keys, like

```py
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "click",
#     "pyjwt",
# ]
# ///

import base64
import click
import jwt
import time
from datetime import datetime, timedelta

@click.command()
@click.argument("b64_key")
def generate_tokens(b64_key):
    """
    Generate ANON_KEY and SERVICE_KEY using a base64-encoded secret key.
    """
    signing_key = base64.b64decode(b64_key)

    now = int(time.time())
    future = now + int(timedelta(days=365.25 * 100).total_seconds())  # 1000 years

    def make_token(role: str) -> str:
        payload = {
            "role": role,
            "iss": "supabase",
            "iat": now,
            "exp": future
        }
        headers = {"alg": "HS256", "typ": "JWT"}
        return jwt.encode(payload, signing_key, algorithm="HS256", headers=headers)

    print("ANON_KEY =", make_token("anon"))
    print("SERVICE_KEY =", make_token("service_role"))

if __name__ == "__main__":
    generate_tokens()

```

malko:
@maelp for the embed key they are the defaults one from the docker compose proposed by supabase, replacing it with random values generated by dokploy don't work here. I don't know how to add a script to a template, but yes it would be a preferable option as I'm not fond of putting in default keys that users may forget to change.  If there's a way to add such script I would be really interested if you can point me in the right direction on how to do it (didn't find any doc on that topic).

Supabase just released a new version, I have not enough time to try to get it to work and test it properly for now. So for now i'll close this pull request and wait for either some more time on my side or someone to take over with the new release.

maelp:
We can ask @Siumauricio 

I guess for now the best is to use the template.yml and add a "mount file" with the content of the script, and mount it in the container, then run it

Siumauricio:
what kind of script would you like to add? @malko 
so I can see how to integrate it in some way.

PR #48, opened 2025-03-29
fix: checkmate template network issues and env vars for emails

mohabgabber:


blueprints/checkmate/docker-compose.yml
```diff
@@ -4,39 +4,49 @@ services:
...
     environment:
       UPTIME_APP_API_BASE_URL: "http://${DOMAIN}/api/v1"
-    ports:
-      - 80
-      - 443
     depends_on:
       - server
+    networks:
+      - dbNet
+      - dokploy-network
```
Siumauricio: Is not required to use dokploy-network for templates, also what's the reason on dbNet? by default they are already connected

Siumauricio (commented):
check

mohabgabber:
> check

Fixed

Siumauricio (approved):

PR #65, opened 2025-04-06
♻️ Refacto minio template

malko:
- add random password generation
- forward port 9000 and expose 9001 through traefik
- add some comments to add some volume setting info

### DIFF
--- a/blueprints/minio/docker-compose.yml
+++ b/blueprints/minio/docker-compose.yml
@@ -3,11 +3,24 @@ services:
   minio:
     image: minio/minio
     volumes:
+      # by default, the MinIO container will use a volume named minio-data
+      # to store its data. This volume is created automatically by Docker.
+      # If you want to use a local directory instead, uncomment the line below
+      # and specify the path to your local directory.
+      # (be warned that ../files is pointing to a subdirectory of /etc/dokploy/compose in dokploy)
+      # - ../files/minio-data:/data
+      # if you uncommented the line above, comment the line below and the volumes section at the end
       - minio-data:/data
     environment:
-      - MINIO_ROOT_USER=minioadmin
-      - MINIO_ROOT_PASSWORD=minioadmin123
+      - MINIO_ROOT_USER=${MINIO_ROOT_USER}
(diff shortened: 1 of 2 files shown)


blueprints/minio/docker-compose.yml
```diff
@@ -3,11 +3,24 @@ services:
...
       - minio-data:/data
     environment:
-      - MINIO_ROOT_USER=minioadmin
-      - MINIO_ROOT_PASSWORD=minioadmin123
+      - MINIO_ROOT_USER=${MINIO_ROOT_USER}
+      - MINIO_ROOT_PASSWORD=${MINIO_ROOT_PASSWORD}
+      - MINIO_BROWSER_REDIRECT_URL=${MINIO_BROWSER_REDIRECT_URL}
     command: server /data --console-address ":9001"
+    ports:
+      - 9000:9000 # expose MinIO api on port 9000
```
Siumauricio: why do we need to expose the api? why we don't use a domain instead?

Siumauricio (changes requested):

blueprints/minio/docker-compose.yml
malko: perhaps I'm getting it wrong but trying to call minio api without a port specified lead to errors (you can try it by adding it as an s3 destination of dokploy). So it seems to me the obvious way to workaround the error. 
I'm eager to learn if there is better way to go.

malko: I finally kept the port to expose the api and remove the domain config for 9000 port.
This solve a problem where pointing the same domain to 9000 and 9001 will sometimes direct the browser to the api port instead of the console port this will then redirect to console but as it is the same domain... it will enter a redirect loop or display XML in the browser instead of showing the console. 
Since port 9000 is exposed outside of traefik, I don't encounter any more problem with this version of minio template

malko:
I also added ```MINIO_BROWSER_REDIRECT=false``` as it seems to sometimes lead to redirect loop under traefik.

Siumauricio (approved):

PR #76, opened 2025-04-13
⬆️ Update Supabase template to version 1.25.03

### LINKED ISSUE #73: [Template update]: Supabase
Hello, Supabase make a new release with many improvements: https://github.com/supabase/supabase/releases/tag/1.25.03

Please implements last version to Dokploy template database.

malko:
- Upgrade to supabase 1.25.03
- Add supavisor which wasn't part of previous template
- Fix mail are now correctly sent (was not the case in previous PR)
- Fix #73

### DIFF
--- a/blueprints/supabase/template.toml
+++ b/blueprints/supabase/template.toml
@@ -1,64 +1,146 @@
 [variables]
 main_domain = "${domain}"
 postgres_password = "${password:32}"
-jwt_secret = "${base64:32}"
 dashboard_password = "${password:32}"
 logflare_api_key = "${password:32}"
-anon_key = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.ewogICJyb2xlIjogImFub24iLAogICJpc3MiOiAic3VwYWJhc2UiLAogICJpYXQiOiAxNzQxNTAwMDAwLAogICJleHAiOiAxODk5MjY2NDAwCn0.muKe0Nrvkf5bMyLoFqAuFypRu3jHAcTYU08SYKrgRQo"
-service_role_key = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.ewogICJyb2xlIjogInNlcnZpY2Vfcm9sZSIsCiAgImlzcyI6ICJzdXBhYmFzZSIsCiAgImlhdCI6IDE3NDE1MDAwMDAsCiAgImV4cCI6IDE4OTkyNjY0MDAKfQ.1KoSiJVueKJNkF59uc84BLqk7h8VdAoVp6Gozqr_vGc"
 
 [[config.domains]]
 serviceName = "kong"
 port = 8_000
 host = "${main_domain}"
 
-[config.env]
-SUPABASE_HOST = "${main_domain}"
-POSTGRES_PASSWORD = "${postgres_password}"
-JWT_SECRET = "${jwt_secret}"
-ANON_KEY = "${anon_key}"
-SERVICE_ROLE_KEY = "${service_role_key}"
-DASHBOARD_USERNAME = "supabase"
-DASHBOARD_PASSWORD = "${dashboard_password}"
-POSTGRES_HOSTNAME = "db"
-POSTGRES_DB = "postgres"
-POSTGRES_PORT = "5432"
-KONG_HTTP_PORT = "8000"
-KONG_HTTPS_PORT = "8443"
-PGRST_DB_SCHEMAS = "public,storage,graphql_public"
-ADDITIONAL_REDIRECT_URLS = ""
-JWT_EXPIRY = "3600"
-DISABLE_SIGNUP = "false"
-MAILER_URLPATHS_CONFIRMATION = '"/auth/v1/verify"'
-MAILER_URLPATHS_INVITE = '"/auth/v1/verify"'
-MAILER_URLPATHS_RECOVERY = '"/auth/v1/verify"'
-MAILER_URLPATHS_EMAIL_CHANGE = '"/auth/v1/verify"'
-ENABLE_EMAIL_SIGNUP = "true"
-ENABLE_EMAIL_AUTOCONFIRM = "false"
-SMTP_ADMIN_EMAIL = "admin@example.com"
-SMTP_HOSTNAME = "supabase-mail"
-SMTP_PORT = "2500"
-SMTP_USER = "fake_mail_user"
-SMTP_PASS = "fake_mail_password"
-SMTP_SENDER_NAME = "fake_sender"
-ENABLE_ANONYMOUS_USERS = "false"
-ENABLE_PHONE_SIGNUP = "true"
-ENABLE_PHONE_AUTOCONFIRM = "true"
-STUDIO_DEFAULT_ORGANIZATION = "Default Organization"
-STUDIO_DEFAULT_PROJECT = "Default Project"
-STUDIO_PORT = "3000"
-IMGPROXY_ENABLE_WEBP_DETECTION = "true"
-FUNCTIONS_VERIFY_JWT = "false"
-LOGFLARE_LOGGER_BACKEND_API_KEY = "${logflare_api_key}"
-LOGFLARE_API_KEY = "${logflare_api_key}"
-DOCKER_SOCKET_LOCATION = "/var/run/docker.sock"
-GOOGLE_PROJECT_ID = "GOOGLE_PROJECT_ID"
-GOOGLE_PROJECT_NUMBER = "GOOGLE_PROJECT_NUMBER"
+[config]
+env = [
+'############',
+'# Secrets',
+'# YOU MUST CHANGE THESE BEFORE GOING INTO PRODUCTION',
+'# https://supabase.com/docs/guides/self-hosting/docker#securing-your-services',
+'############',
+'',
+'SUPABASE_HOST=${main_domain}',
+'POSTGRES_PASSWORD=${postgres_password}',
+'JWT_SECRET=your-super-secret-jwt-token-with-at-least-32-characters-long',
+'ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyAgCiAgICAicm9sZSI6ICJhbm9uIiwKICAgICJpc3MiOiAic3VwYWJhc2UtZGVtbyIsCiAgICAiaWF0IjogMTY0MTc2OTIwMCwKICAgICJleHAiOiAxNzk5NTM1NjAwCn0.dc_X5iR_VP_qT0zsiyj_I_OZ2T9FtRU2BBNWN8Bu4GE',
+'SERVICE_ROLE_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyAgCiAgICAicm9sZSI6ICJzZXJ2aWNlX3JvbGUiLAogICAgImlzcyI6ICJzdXBhYmFzZS1kZW1vIiwKICAgICJpYXQiOiAxNjQxNzY5MjAwLAogICAgImV4cCI6IDE3OTk1MzU2MDAKfQ.DaYlNEoUrrEn2Ig7tqibS-PHK5vgusbcbo7X36XVt4Q',
+'DASHBOARD_USERNAME=supabase',
+'DASHBOARD_PASSWORD=${dashboard_password}',
+'SECRET_KEY_BASE=UpNVntn3cDxHJpq99YMc1T1AQgQpc8kfYTuRgBiYa15BLrx8etQoXz3gZv1/u2oq',
+'VAULT_ENC_KEY=your-encryption-key-32-chars-min',
+'',
+'',
+'############',
+'# Database - You can change these to any PostgreSQL database that has logical replication enabled.',
+'############',
+'',
+'POSTGRES_HOST=db',
+'POSTGRES_DB=postgres',
+'POSTGRES_PORT=5432',
+'# default user is postgres',
+'',
+'',
+'############',
+'# Supavisor -- Database pooler',
+'############',
+'POOLER_PROXY_PORT_TRANSACTION=6543',
+'POOLER_DEFAULT_POOL_SIZE=20',
+'POOLER_MAX_CLIENT_CONN=100',
+'POOLER_TENANT_ID=your-tenant-id',
+'',
+'',
+'############',
+'# API Proxy - Configuration for the Kong Reverse proxy.',
+'# Following ports should not be changed for a dokploy config unless you know what you are doing.',
+'############',
+'',
+'KONG_HTTP_PORT=8000',
+'KONG_HTTPS_PORT=8443',
+'',
+'',
+'############',
+'# API - Configuration for PostgREST.',
+'############',
+'',
+'PGRST_DB_SCHEMAS=public,storage,graphql_public',
+'',
+'',
+'############',
+'# Auth - Configuration for the GoTrue authentication server.',
+'############',
+'',
+'## General',
+'SITE_URL=http://localhost:3000',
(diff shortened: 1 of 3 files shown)


malko:
@maelp if you want to iterate on this to add generation script for the jwt and keys it would be super cool :). I think you can even propose it to upstream supabase repository.

Siumauricio:
I can add a new helper similar to the ones we have https://github.com/Dokploy/templates?tab=readme-ov-file#templatetoml-structure that is specifically to generate the jwt and keys, what I don't know is what implementation is needed to generate those jwts that supabase needs? https://github.com/Dokploy/dokploy/blob/canary/packages/server/src/templates/processors.ts#L68

maelp:
@Siumauricio would be nice indeed! Perhaps have a way to configure eg ${jwt:sha-256:RSA} etc

It's a standard JWT, I think there are many nodejs libs for that, here are a few from a quick Google search (haven't tested them):
- https://github.com/auth0/node-jsonwebtoken (from Auth0)
- https://github.com/jwtk/njwt

malko:
@Siumauricio perhaps allowing the jwt helpers to take some more parameters should do the trick. we need to pass in some json payload, the encode key and the algorythm.
Here a sample of what it could look like in toml config file
```
jwtSecret=${password:32}
anonkey_payload={
  role = 'anon",
  iss= "supabase",
  iat = 1744495200,  # perhaps another helper to help deal with timestamps
  exp = 1902261600
}
servicekey_payload={
  role = 'service_role",
  iss= "supabase",
  iat = 1744495200,  # perhaps another helper to help deal with timestamps
  exp = 1902261600
}
anonkey=${jwt:"HS256":jwtSecret:anonkey_payload}
servicekey=${jwt:"HS256":jwtSecret:servicekey_payload}
```
what do you think  ?

malko:
In the meantime I think we can still push this version as I've added link to the supabase documentation https://supabase.com/docs/guides/self-hosting/docker#securing-your-services inside the generated env file so people can easily find the link to generate those token directly while reviewing the file.

maelp:
the helper could use the iat to be Date.now() and expire_at to be defined as parameter

eg ${jwt:secret_key_variable_name:expire_in_delta}

like ${jwt:mysecret:1y} to expire in 1y or something

malko:
@Siumauricio also if you point me in the right direction i can propose a PR for that if you want probably within a week or two.

aizazzaheer:
Great work @malko - any idea when this would be updated on the main Dokploy blueprints template folder? https://github.com/Dokploy/templates/tree/main/blueprints/supabase

Siumauricio:
@malko Here's the code where we process the utilities https://github.com/Dokploy/dokploy/blob/canary/packages/server/src/templates/processors.ts#L68 I think you can easily extend the jwt function, also we have a couple test https://github.com/Dokploy/dokploy/blob/canary/apps/dokploy/__test__/templates/config.template.test.ts#L6

malko:
@Siumauricio will have a look at it this week-end

malko:
@Siumauricio Will update this PR to match changes to jwt helpers introduces in PR https://github.com/Dokploy/dokploy/pull/1749. I will also add some doc to the dolploy/templates readme to reflect the changes.

malko:
latest 2 commits requires PR on dokploy to be merged, so in order for this to reach users here's what needed to be done @aizazzaheer :
- PR for new helpers needs to be approved on Dokploy/dokploy repo
- A new dokploy release has to be made with the aforementioned PR merged
- This PR get merged
- users can starts to use it

This leaves me with some question about releasing a template that will require specific version of dokploy to be used. I don't know how this is handle yet or if it is handled at all.  
- Perhaps we should start to include some data in meta.json to add a minimal dokploy version required for templates.
- How can we have 2 version of a template for different dokploy version ?

I think supabase template is particularly awaited by users (maybe i'm wrong), so we should perhaps release it without such strong requirement as latest dokploy version. what do you think @Siumauricio

Siumauricio:
Hey I have already released version 0.21.8 there should be the jwt changes, thanks for the implementation. 

I think it would be a good idea to add some field as you mentioned based on the minimum version of dokploy, to prevent people trying to create templates with old versions of dokploy.


in this case you can simply create another template with the same name and some prefix with the version name or something like that to differentiate them.

malko:
> in this case you can simply create another template with the same name and some prefix with the version name or something like that to differentiate them.

Something like pre0.21.8-supabase and some addition in the descritpion ?

malko:
thinking about that version story, I think if the templates project could use the helpers directly, it would be able to generate version of the template with helpers already applied and so we wouldn't have to depend on dokploy versions. 

Another way if we don't want to have helpers in this projects would probably to handle a queryString parameter with the target version of dokploy, so we can filter templates on this in the future.

kimtiago:
I'm waiting for this PR so much

malko:
@useranik12 can you please provide more details, I didn't had any issues on my test env. Do you have some logs ? I personally don't uses storage so perhaps I missed something. 

@kimtiago  you can try this by visiting https://malko-supabase-1-25-03.templates-70k.pages.dev/ and grab the base64 encoded config. Pay attention to the dokploy version you use, there's 2 different version depending your dokploy version.

ArtuArtaza:
Where I have to put the base64 to work?

malko:
@ArtuArtaza create a new docker-compose service go to the "advanced" tab and paste it in the "import" field

Siumauricio:
I tested and it works fine, the only thing that is not working is the storage as @useranik12, I attach a video of the error
https://github.com/user-attachments/assets/b4ec79f3-8d2b-454c-8fcf-2c5d15b1a574

you probably have to manually adjust the volumes of some script, or else manually enable some permissions in the dashboard.

https://github.com/orgs/supabase/discussions/6765 related

franciscoesser:
Hi @malko, great work on this!

I wanted to add a couple of issues we've encountered while using Supabase with Dokploy. Hopefully, these findings will help optimize your contribution:

1. **Inability to create multiple Supabase instances on a single server:** When attempting to deploy a second instance, the process fails because container names conflict with the existing installation. We've tried randomizing names, but the issue persists.

2. **DB Backup Restoration Errors:** Restoring backups using the new DB backup functionality results in errors. This seems to be related to how Supabase manages different users and roles within its database tables, causing conflicts during restoration.

Thanks again for your contribution!

malko:
> I tested and it works fine, the only thing that is not working is the storage as @useranik12, I attach a video of the error https://github.com/user-attachments/assets/b4ec79f3-8d2b-454c-8fcf-2c5d15b1a574
> 
> you probably have to manually adjust the volumes of some script, or else manually enable some permissions in the dashboard.
> 
> https://github.com/orgs/supabase/discussions/6765 related

@Siumauricio For what I understand this is more related to how you configure RLS in supabase that the template working or not, unless i misanderstood something ?

malko:
@useranik12 if you are running recent version of dokploy >= 0.21.8 and the appropriate template (for dokploy >= 0.21.8) you should have jwt and secrets automatically generated for you.

By the way I will update the PR soon with version 1.25.04 of supabase.

malko:
@useranik12 also forgot to say that edge function are working, but you need the cli to deploy function, there's no GUI in self-hosted version. There's some limitation on the features on self-hosted supabase look at the doc they are well explained. And for AI you can add your open ai key in the env variables

malko:
@useranik12 I'm not an official dokploy maintainer, just a contributor, don't hesitate to propose feature or error  in the issue section of this repository. This thread is only about the supabase template. And why not propose PR yourself ;)

Siumauricio:
> OK So now it's updated to latest 1.25.04 version:
> 
> * Work around logs => I had hard time getting logs to work because config is hard coded and use container names to work. I  added some dynamic container names and handle them in the vector.yml file with a replace => this is not the cleaniest way but it works. Thus it should ease the configuration of multiple supabase instance on a single host.
> * I've addded a bunch of explanation on which variables should be checked in order to have a working install + link to official supabase self hosting documentation.
> * Regarding the storage problem mentionned earlier, I had no issue with this version in adding a new bucket and upload files to it. But as @Siumauricio pointed out in previous comments it may be related to configuration of policies inside of supabase. I don't think there's an issue with the actual template regarding this.
> 
> Any help on this would be greatly appreciated if more modification are required, supabase is not the easiest service to get running to be honnest. If some of you are supabase gurus 🙂, I'm eager on any advice to make things better. Hope this is the good one this time ;)

I see you added the container_name, is there any chance you can remove that?

I guess you need to access the container name, however there is no deterministic way.


I think I could add the environment variable ${appName} which is used to generate the projectName from docker compose https://docs.docker.com/compose/how-tos/project-name/, that way you would have to call for example inside the file mount something like ${appName}_{serviceName}_1 

image

What do you think about this? I believe that will work

malko:
@sslyxhz => can you provide your env settings I'm pretty sure this is due to missing configuration of the env variables

malko:
@Siumauricio 

> I see you added the container_name, is there any chance you can remove that?
No this is required (it was already the case in the actual supabase template)
> I guess you need to access the container name, however there is no deterministic way.
Yes this is required
> I think I could add the environment variable ${appName} which is used to generate the projectName from docker compose https://docs.docker.com/compose/how-tos/project-name/, that way you would have to call for example inside the file mount something like ${appName}_{serviceName}_1

> What do you think about this? I believe that will work
Not sure, probably we need to try it out it will require change to the mounted file: /volumes/logs/vector.yml
or we can probably just change the ```container_name_prefix``` in the template file to use that appname
```
container_name_prefix = "supabase-${appName}" 
```
This should work with minimal modification. I can't test until the current canary version is deployed, i've no more testing environment to install the canary version on (all my servers running the latest stable version)

blueprints/pre0.21.8-supabase/template.toml
```diff
@@ -0,0 +1,1069 @@
...
+filePath = "/volumes/logs/vector.yml"
+content = """api:
+  enabled: true
+  address: 0.0.0.0:9001
+
+sources:
+  docker_host:
+    type: docker_logs
+    exclude_containers:
+      - ${container_name_prefix}-vector
```
Siumauricio: I mean but having the appName, we can easily know which container it refers to within each volume for example something like this 

```suggestion
      - ${APP_NAME}-vector_1
```

this should work, if I'm not wrong

malko: We also need more than that: for logs to appear in supabase they require the correct .appname in the vector.yml
```
      kong: '.appname == "supabase-kong"'
      auth: '.appname == "supabase-auth"'
      rest: '.appname == "supabase-rest"'
      realtime: '.appname == "realtime-dev.supabase-realtime"'
      storage: '.appname == "supabase-storage"'
      functions: '.appname == "supabase-edge-functions"'
      db: '.appname == "supabase-db"'
```
first because they are filtered by those names, but also because inside supabase studio they won't display anything with a different name than that even if you changes this settings to the correct container name. 

So to make it work I've added a replace : ```.appname = replace!(del(.container_name), "${container_name_prefix}", "supabase")``` (wich i'm not fan of as it adds small overhead on each log received)

This also mean we should not add suffix like in your example: ```${APP_NAME}-vector_1``` unless we also remove it. 
So we potentially can replace this with something like: ```.appname = replace!(del(.container_name), "${APP_NAME}", "supabase")``` but in the end it is mandatory for logs to appear in supabase studio to use this names in the log themselves (ie: supabase-kong, supabase-auth, supabase-rest...).

I really think that the most simple solution is to simply change the ```container_name_prefix``` to include the ```${APP_NAME}``` as proposed earlier.

malko: @Siumauricio we can perhaps just reverse the prefix like this: 
```
[variables]
main_domain = "${domain}"
...
container_name_prefix = "${appName}-supabase" 
```
This will keep more meaningful names, and I think it is easy to lookup in the env file the CONTAINER_PREFIX  value to see correlation within the docker-compose.yml

sslyxhz:
> @sslyxhz => can you provide your env settings I'm pretty sure this is due to missing configuration of the env variables

I've solved the problem by running the following command and redeploying it. 
`sudo rm -rf docker/volumes/db/data`

Thank you for your work!

malko:
@useranik12 this is not related to the template but rather a supabase config you need to do a quick google search lead to https://github.com/orgs/supabase/discussions/20241 which I suppose is the problem you are facing off.

blueprints/pre0.21.8-supabase/template.toml
malko: @Siumauricio do you want me to make the change in the container_name_prefix. I think it would be really nice to get that template out shortly :)

Siumauricio: Yes @malko, please, with those changes I merge! sorry for the delay!

malko: @Siumauricio hey, we all have a life :) no problem for the delay. I've just updated the PR I think it should be good :)

blueprints/supabase/template.toml
```diff
@@ -1,64 +1,179 @@
...
 postgres_password = "${password:32}"
-jwt_secret = "${base64:32}"
 dashboard_password = "${password:32}"
 logflare_api_key = "${password:32}"
-anon_key = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.ewogICJyb2xlIjogImFub24iLAogICJpc3MiOiAic3VwYWJhc2UiLAogICJpYXQiOiAxNzQxNTAwMDAwLAogICJleHAiOiAxODk5MjY2NDAwCn0.muKe0Nrvkf5bMyLoFqAuFypRu3jHAcTYU08SYKrgRQo"
-service_role_key = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.ewogICJyb2xlIjogInNlcnZpY2Vfcm9sZSIsCiAgImlzcyI6ICJzdXBhYmFzZSIsCiAgImlhdCI6IDE3NDE1MDAwMDAsCiAgImV4cCI6IDE4OTkyNjY0MDAKfQ.1KoSiJVueKJNkF59uc84BLqk7h8VdAoVp6Gozqr_vGc"
+secret_key_base = "${password:64}"
+vault_enc_key = "${password:32}"
+jwt_secret = "${password:32}"
+container_name_prefix = "${appName}-supabase"
```
Siumauricio:
```suggestion
container_name_prefix = "${APP_NAME}-supabase"
```

blueprints/pre0.21.8-supabase/template.toml
```diff
@@ -0,0 +1,1069 @@
+[variables]
+main_domain = "${domain}"
+postgres_password = "${password:32}"
+dashboard_password = "${password:32}"
+logflare_api_key = "${password:32}"
+container_name_prefix = "${appName}-supabase"
```
Siumauricio:
```suggestion
container_name_prefix = "${APP_NAME}-supabase"
```

Siumauricio (changes requested):

franciscoesser:
Hi, we’re also having issues. 

We’re using the latest version of Dokploy with the new Supabase template in a clean server. 

Users and buckets can’t be created; it displays the same error.

Thanks for your time and effort!

1CleanShot 2025-05-28 at 09 26 26@2x

malko:
Have you set policies inside supabase ?

franciscoesser:
Hi @malko, it's a totally clean installation.

- Clean Ubuntu Server.
- Dokploy (0.22.7).
- Supabase template (1.25.04 / dokploy >= 0.22.5).
- No policies, no customization.

I tried with the default anon and service keys and generating new ones from the Supabase website, but still failing.

What do you recommend? Thank you!

malko:
@franciscoesser first try to set some policies in supabase. You can also send me your env file without secrets I will have a look. Also check the domain is correctly set in the env file and the domain tab settings.

PR #81, opened 2025-04-18
Home Assistant template

gentslava:
Снимок экрана 2025-04-18 в 11 48 12
Снимок экрана 2025-04-18 в 11 50 09


Siumauricio:
Why the domain isnt working when I try to access ?

gentslava:
> Why the domain isnt working when I try to access ?

1) First launch about 30 seconds.
2) When you get 400 Bad request. There is a Home Assistant restriction for reverse proxy. You have to configure it in configuration.yaml (via SSH for example)
https://www.home-assistant.io/integrations/http/#reverse-proxies
Снимок экрана 2025-04-21 в 12 56 25

gentslava:
@Siumauricio updated compose.
1) Absolute path for HA data. To avoid overwriting files during restart.
2) Updated ports section.

Siumauricio:
Ok, is it possible to mount that configuration file? so that the user does not have to create it manually? even if he has to do some manual configuration but at least it has the base structure so that the user can add his ip, etc. @gentslava

gentslava:
> Ok, is it possible to mount that configuration file? so that the user does not have to create it manually? even if he has to do some manual configuration but at least it has the base structure so that the user can add his ip, etc. @gentslava

Sure, great idea! I'll make this change.

gentslava:
@Siumauricio done

Siumauricio (approved):

PR #84, opened 2025-04-19
feat: add Qdrant template

thearnabmondal:
This PR adds a new template for [Qdrant](https://qdrant.tech/) – a blazing-fast vector search engine, perfect for powering semantic search, LLMs, and AI-driven apps.

### 🏷 Tags
`vector-db`, `database`, `search`, `AI-ready`

### DIFF
--- a/blueprints/qdrant/docker-compose.yml
+++ b/blueprints/qdrant/docker-compose.yml
@@ -1,15 +1,14 @@
-version: "3.8"
 services:
   qdrant:
     image: qdrant/qdrant:latest
     restart: unless-stopped
     ports:
-      - "6333"
-      - "6334"
+      - "6333:6333"
     volumes:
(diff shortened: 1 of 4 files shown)


thearnabmondal:
Tried my best, but custom domain still isn’t working – though I did get it working via `IP:PORT`.

Previously, I deployed Qdrant without Dokploy using a super simple `docker-compose.yml` with just:

```yaml
expose:
  - 6333
```

…and Caddy as the reverse proxy, and it worked fine.

If anyone has experience getting custom domains working in Dokploy for services like this, would love a hand

Siumauricio:
It seems that the service is being exposed on localhost instead of 0.0.0.0.0, which will cause the routing not to work, this seems to be a qdrant problem.

PR #110, opened 2025-05-02
Update Listmonk to version 5.0.0

jeff-h:
Also move environment variables to Dokploy's environment management.


jeff-h:
This needs a little more work.

PR #115, opened 2025-05-04
Fix: Bump version of Formbricks template

### LINKED ISSUE #114: Formbricks template doesn't work
The formbricks template doesn't work out of the box

driaug:
This fix bumps the version of the Formbricks template to `3.1.5` from `3.1.3` to resolve #114 

The root cause of the issue is documented here: https://github.com/formbricks/formbricks/issues/4704

### DIFF
--- a/blueprints/formbricks/docker-compose.yml
+++ b/blueprints/formbricks/docker-compose.yml
@@ -22,7 +22,7 @@ services:
 
   formbricks:
     restart: always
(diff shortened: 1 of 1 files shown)


Siumauricio:
Thanks

maelp:
Works nicely! I think their last version is 3.9.0 don't know if there is already an image for this

PR #144, opened 2025-05-17
Update OpenResty Manager

Safe3:
Release version 1.2.0 and update docker-compose.yml  "network_mode: host" to support custom listen on any ports


Siumauricio:
I don't know if you have tried to deploy but this will not work if you add a network_mode....

Safe3:
OK, I already removed "network_mode: host", since it does't work on this platform, sorry for this!

Siumauricio:
Yes it works, but you have to manually add the domains, because when you add a domain in dokploy it automatically adds a network which causes an incompatibility with docker, however if you remove the domains you can use that option.

Siumauricio (approved):

PR #149, opened 2025-05-21
FreeScout template

aropsta:
This PR adds a FreeScout template to Dokploy Templates.

FreeScout is an open-source help desk and shared inbox system, providing a self-hosted alternative to services like HelpScout and Zendesk.

Features included in this template:
- Support for external MySQL/MariaDB databases
- Persistent storage for all data
- Admin user configuration
- Automatic HTTPS support via Dokploy

Testing Status:
- The template has been thoroughly tested with external MySQL database
- Installation completes successfully and the application is accessible
- Admin user creation works properly
- Web interface functions as expected

Important Notes:
- FreeScout requires a MySQL/MariaDB database (PostgreSQL is not supported)
- The database user needs full permissions on the database
- The application will automatically set up the fresh database tables on first run


aropsta:
Although I've tested locally to ensure everything works, when I import the template using the Preview URL generated code, the deployment gets various issues. 

- It get stuck at copying files in the logs. I've checked to make sure that it wasn't just a long operation. It really does get stuck and doesn't move forward. Restarting the container makes it proceed but it gets stuck at another stage again.

When I deploy directly using using a new docker compose or stack service It successfully deploys and logs do not get stuck.

aropsta:
It's done. Everythings working. I faced some issues with trying to use an external (dokploy network) database but decided to just keep it simple and use the internally deployed database. 

Must be deployed as docker compose, NOT docker stack.

Siumauricio (approved):

PR #151, opened 2025-05-24
feat(blueprint): healthcheck added for rybbit backend service

nktnet1:
From PR: https://github.com/rybbit-io/rybbit/pull/261

### DIFF
--- a/blueprints/rybbit/docker-compose.yml
+++ b/blueprints/rybbit/docker-compose.yml
@@ -36,14 +36,14 @@ services:
       - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
     volumes:
       - postgres_data:/var/lib/postgresql/data
-    restart: unless-stopped
-    networks:
-      - internal
     healthcheck:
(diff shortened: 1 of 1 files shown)


blueprints/rybbit/docker-compose.yml
```diff
@@ -36,14 +36,14 @@ services:
...
-    networks:
-      - internal
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}"]
       interval: 30s
       timeout: 10s
       retries: 3
+    restart: unless-stopped
+    networks:
+      - internal
```
Siumauricio, line 46: why do we need this internal network ?

nktnet1: I don't think it's necessary, it's just what's in the guide:
- https://www.rybbit.io/docs/self-hosting-advanced#docker-compose-setup-with-npm-without-exposing-ports

I'd imagine the reason is so that services like the Client (frontend) and Caddy would not be connected to Postgres and Clickhouse, as they're linked by web, whereas the server (backend) is linked to both web and internal as the bridge.

PR #153, opened 2025-05-25
fix(update): use caddy built-in metrics endpoint for health check

rexwangcc:
## Why

With the current health check, we'd get tons of warning logs under the `Caddy` container complaining something like:

```json
{"level":"warn","ts":1748150819.4112387,"logger":"http.handlers.reverse_proxy","msg":"aborting with incomplete response","upstream":"rybbit_client:3002","duration":0.004111297,"request":{"remote_ip":"127.0.0.1","remote_port":"49632","client_ip":"127.0.0.1","proto":"HTTP/1.1","method":"GET","host":"rybbit_client:3002","uri":"/","headers":{"X-Real-Ip":[""],"Via":["1.1 Caddy"],"Accept":["/"],"X-Forwarded-For":["127.0.0.1"],"User-Agent":["Wget"],"X-Forwarded-Proto":["http"],"X-Forwarded-Host":["127.0.0.1:80"]}},"error":"writing: write tcp 127.0.0.1:80->127.0.0.1:48632: write: broken pipe"}

```

it is because for the current health check `["CMD", "wget", "--spider", "--quiet", "http://127.0.0.1:80"]`, the `--spider` flag tells wget to not download the response body - it just checks if the URL exists and then immediately closes the connection; however, `Rybbit` Next app is trying to send the full response, but wget has already closed the connection, causing the "broken pipe" error.

## How

- Ideally we would use `curl` to perform the health checks, but `caddy` image does not come with it
- Looking at `caddy_clickhouse`, and docker compose's conventions, apparently we want to scope the health checks to its own service rather than let it cascade, so I changed the health check to only probe `caddy`'s built-in `/metrics` endpoint
- using `--no-verbose` over `--quiet` for logging purposes, former does not suppress errors.


rexwangcc:
@Siumauricio sorry for the bump if i tagged the wrong person, but i can't find a way to request for reviews 😅

Siumauricio (approved):

PR #135, opened 2025-05-14
feat(blueprint): added seafile

nktnet1:
Added Seafile, an open source cloud storage system for file sync, share and document collaboration.

- Github: https://github.com/haiwen/seafile
- Website: https://seafile.com
- Documentation: https://manual.seafile.com/12.0

Resolves #32.

PR #160, opened 2025-06-01
fix(blueprint): rybbit v1.2.0

nktnet1:
Changes:
- remove /api stripping
- remove caddy reverse proxy
- added clickhouse config in volume mount

Release notes:
- https://github.com/rybbit-io/rybbit/releases/tag/v1.0.0

Migration guide:
- https://www.rybbit.io/docs/v1-migration

### DIFF
--- a/blueprints/rybbit/docker-compose.yml
+++ b/blueprints/rybbit/docker-compose.yml
@@ -6,11 +6,11 @@
 #
 # They should share the same base URL for cookies
 # to function correctly. The `BASE_URL` environment
-# variable should also be set accordingly.
+# variable should also be set accordingly for CORs.
 
 services:
   rybbit_clickhouse:
-    image: clickhouse/clickhouse-server:25.4.2
+    image: clickhouse/clickhouse-server:25.5.1
     volumes:
       - clickhouse_data:/var/lib/clickhouse
       - clickhouse_config:/etc/clickhouse-server/config.d
@@ -37,7 +37,7 @@ services:
       - internal
 
   rybbit_postgres:
-    image: postgres:17.4
+    image: postgres:17.5
     environment:
       - POSTGRES_DB=${POSTGRES_DB}
       - POSTGRES_USER=${POSTGRES_USER}
--- a/blueprints/rybbit/docker-compose.yml
+++ b/blueprints/rybbit/docker-compose.yml
@@ -67,7 +67,7 @@ services:
       rybbit_postgres:
         condition: service_started
     healthcheck:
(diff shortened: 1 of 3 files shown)


nktnet1:
@Siumauricio can you help test Rybbit's v1.0.0 migration?

I tried it and everything works only when "isolated compose" is turned off, otherwise I get a 504 Gateway error.

This is similar to the issue I faced with the seafile template (#135), where I get a 504 gateway timeout when accessing the rybbit_backend service.

nktnet1:
> Check the video is working fine https://github.com/user-attachments/assets/2f9b4f3d-c9c7-45b8-9258-60d4fcb8fb1e

Thanks for setting it up @Siumauricio - however, that's only the frontend client right?
Are you able to sign up? I think the issue is more so with the backend.

You could also try visiting
- https://your.domain/api/health

to see if it can be reached.

nktnet1:
That's odd...

I followed your directions in the video and removed all the network configurations, and things are working correctly now, even for sign-up.

I'm using a custom domain though, so will try again with default traefik.

nktnet1:
> Is the service likely to have certain restrictions on proxy and url handling?

not 100% sure. What I have deduced is:
1. Prior to removing the "networks" configuration in docker-compose.yaml (https://github.com/Dokploy/templates/pull/160/commits/0b44930611549dc103171d9c8ace46ce920e1ec5)
    1. isolated deployment **enabled**
        - ❌ backend always unreachable
    1. isolated deployment **disabled**
        - ✅ works with traefik domain (HTTP)
        - ✅ works with custom domain (HTTPS)
       
1. After removing the network configurations (https://github.com/Dokploy/templates/pull/160/commits/0b44930611549dc103171d9c8ace46ce920e1ec5)
    1. isolated deployment **enabled**
        - ⚠️ partially works with traefik domain (HTTP), can create users, but not organizations
        - ✅ works with custom domain (HTTPS)
    1. isolated deployment **disabled**
        - ❌ backend always unreachable
        
Pretty confusing 😕.

nktnet1:
@Siumauricio I got confirmation that Rybbit only supports HTTPS, so issues with traefik is not within scope:
- https://github.com/rybbit-io/rybbit/issues/322#issuecomment-2930745085

I've updated the documentation in 1e79ddf7ada445d881f15b9aebd0dab987f028de.

I think the PR is ready to merge - it was confirmed to be working by another user here:
- https://github.com/rybbit-io/rybbit/issues/322#issuecomment-2930792831

PR #197, opened 2025-06-24
feat: add Livekit Blueprint

koch3092:


Siumauricio:
How did you test the template? When I access to all the domains I get 404 not found, and infinite loading

PR #205, opened 2025-07-03
fix:modify TemplateDialog.tsx to make the dialog support Chinese

siesteven:
Dear author
      I use Chinese to describe   template ,however the browser output an error . Then I found the param of btoa need to be encoded . Now I have wrapped the param and encoded them , to  make sure the function goes well .
      Pls accept my PR
thanks

### DIFF
--- a/app/src/components/TemplateDialog.tsx
+++ b/app/src/components/TemplateDialog.tsx
@@ -55,7 +55,7 @@ const TemplateDialog: React.FC<TemplateDialogProps> = ({
       config: templateFiles.config || "",
     };
 
-    return btoa(JSON.stringify(configObj, null, 2));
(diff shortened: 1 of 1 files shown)


Siumauricio:
This is breaking the import feature from dokploy, so please check again this

nktnet1:
If you use `encodeURIComponent` while exporting, then you should `decodeURIComponent when importing.

Relevant code here:
- https://github.com/Dokploy/dokploy/blob/3ed9da147b47ea4966778cdf189d12d46d42cb9b/apps/dokploy/components/dashboard/application/advanced/import/show-import.tsx#L116

For backward-compatibility reasons, perhaps try as base64 first, and if error, try decoding URI before `atob`.

PR #213, opened 2025-07-06
add casdoor template

thaarrtt:
add casdoor template


blueprints/casdoor/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+      retries: 5
+      start_period: 30s
+
+volumes:
+  casdoor-postgres-data:
+  casdoor-data:
+
+networks:
+  dokploy-network:
+    external: true
```
Siumauricio: Can you apply the new suggestions we've added in the readme File?

https://github.com/Dokploy/templates?tab=readme-ov-file#general-suggestions-when-creating-a-template

Siumauricio (changes requested):

blueprints/casdoor/docker-compose.yml
thaarrtt: done and its applied

PR #298, opened 2025-08-23
Fix Garage S3 storage separation

vniehues:
## What is this PR about?

After testing with multiple instances I noticed that they shared their data and meta directories. 

This PR fixes that by using volumes instead.

---

### Checklist

- [x] I have read the suggestions in the [README.md file](https://github.com/Dokploy/templates?tab=readme-ov-file#general-suggestions-when-creating-a-template)
---


channchetra:
I always got this error and can't login. I was deploy on 3 different Dokploy instant can you help on this? 
image

vniehues:
I published some fixes to the template in https://github.com/Dokploy/templates/pull/367.
When this is merged and you still face this issue, please let me know and we can go through the logs.

PR #312, opened 2025-08-26
feat(blueprint) add kaneo task managment template

mikield:
A template for Kaneo.app

A self-hosted task management solution.


mikield:
I see that validation of meta.json is not passing, but I can not understand why it says that there are 255 entries, and it was 255 entries, but I have added a new one.

PR #315, opened 2025-08-27
fix: appflowy template not including nginx

naterfute:
I made an oversight when designing the appflowy template that I couldn't test because I wasn't using the appflowy app. I have tested the app and it should work without any issues now.


naterfute:
I can't figure out how the mounts work though. So, can I have somebody verify that before merge?

PR #328, opened 2025-09-02
Add NTFY (Item requested on issue #145)

jorgegomzar:
# Description
This PR aims to add NTFY configuration files.

# Context
NTFY was requested here --> https://github.com/Dokploy/templates/issues/145

I have experience self-hosting NTFY already in the cloud, but this is my first contribution to Dokploy templates.

NTFY requires access to HTTPS certificate files in order to use websockets as expected. I noticed some templates already mounted ./certs directory, so I assumed this is how we can expose Dokploy SSL certificates for the containers.
If there is a better way of doing this please let me know and I'll update it ASAP.

Thanks!


jorgegomzar:
Extensively tested in https://github.com/Dokploy/templates/pull/326, decided to start a new clean PR for this feature.

PR #330, opened 2025-09-03
fix: add signoz alertmanager provider in the template

nijat:
Signoz template are missing this environment variable:
`      - SIGNOZ_ALERTMANAGER_PROVIDER=signoz`


nijat:
@thaarrtt can you check this PR as well, without having this, alert manager was not working, default was external alert manager.

PR #332, opened 2025-09-03
Add Java runtime blueprint and metadata

devdamo:
Introduces a new Java blueprint with Docker Compose, template configuration, and logo for multi-version Java environments. Updates meta.json to register the Java runtime, supporting versions 8-21 for use cases like Minecraft servers and Spring Boot apps.


devdamo:
Hi, this is a Java setup, but it’s designed to work with Minecraft. Currently, it fails by default because it looks for a JAR file, which I haven’t added. I’d like to know if I can implement an installation for a Minecraft version instead. If so, should this template be categorized as Minecraft or remain as Java? Please advise.

PR #356, opened 2025-09-10
WGER fitness blueprint

maarteNNNN:


Siumauricio:
I'll close this I don't think it's worth having a PR open for very long.

PR #377, opened 2025-09-18
refactor(invoiceshelf): switch from PostgreSQL to SQLite and simplify configuration

jaainil:
## What is this PR about?

New PR of invoiceshelf

- Removed separate PostgreSQL service and switched to SQLite database
- Renamed app service from 'invoiceshelf-app' to 'webapp'
- Simplified environment variables and removed redundant database configuration
- Added mail configuration options with sensible defaults
- Updated volume and mount configurations for better data persistence
- Removed network configuration as it's no longer needed

## Checklist

- [x] I have read the suggestions in the README.md file https://github.com/Dokploy/templates?tab=readme-ov-file#general-suggestions-when-creating-a-template

### DIFF
--- a/blueprints/invoiceshelf/docker-compose.yml
+++ b/blueprints/invoiceshelf/docker-compose.yml
@@ -1,57 +1,30 @@
 version: "3.8"
 
 services:
-  invoiceshelf-postgres:
-    image: postgres:15
-
-    volumes:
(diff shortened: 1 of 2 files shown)


Siumauricio:
Hey, thank you for the PR, in this case I prefer to have PostgresSQL since we can have backups, sqlite only volume backups, can you make a new version for this template?

Siumauricio:
Close due to inactivity

PR #391, opened 2025-09-22
feat: add Odoo 19 blueprint

rejamen:
- [x] I have tested in a VPS using the Base64 generated on the PR preview and its working without any issue. I had Odoo version 19.0 running 
- [ ] If this PR is ok like this, and merged I can open two more for the latest supported Odoo versions 18.0 and 17.0 as well


Siumauricio:
@rejamen Thanks for the PR, makes sense i'm open to receive different PRs from different versions

PR #407, opened 2025-09-30
feat: add Odoo 18 blueprint

rejamen:
It make sense for the Odoo world to have at least the 3 latest supported version from Odoo. 
This is why we need this PR with v18.0

- [x] Tested on my VPS with the base64 preview and it works.


meta.json
```diff
@@ -3620,7 +3635,7 @@
       "docs": "https://www.odoo.com/documentation/"
     },
     "tags": [
-      "crm"
+      "erp"
```
rejamen, line 3638: This would be the more accurate tag

PR #410, opened 2025-10-01
Add mealie template

lwndp:
Mealie is a recipe tracking and meal planner app


lwndp:
When importing the Base64 template my dokploy instance throws "error processing template". Can you see why?

PR #432, opened 2025-10-06
chore: update Tolgee to latest version and fix SMTP config typo

dzienisz:
Update Tolgee Template to Latest Version & Fix Configuration

Changes
- **Version**: Updated from v3.80.4 → latest for auto-updates
- **Env Variables**: Fixed TOLGEE_SMTPHOST → TOLGEE_SMTP_HOST

### DIFF
--- a/blueprints/tolgee/docker-compose.yml
+++ b/blueprints/tolgee/docker-compose.yml
@@ -1,5 +1,6 @@
 version: "3"
 
+
 services:
   app:
     image: tolgee/tolgee:latest


meta.json
```diff
@@ -4854,7 +4854,7 @@
   {
     "id": "tolgee",
     "name": "Tolgee",
-    "version": "v3.80.4",
+    "version": "latest",
```
JanCizmar, line 4857: Is the latest pulled when user requires the pull or is it pulled automatically?

if it's automatically, this might be a potential issue.

Harikrishnan1367709:
Screenshot 2025-10-07 at 8 57 19 PM
Screenshot 2025-10-07 at 8 57 37 PM

meta.json
dzienisz: @Siumauricio @maelp can you help me with this question? What if we release a breaking change? They will automatically get the latest, and it will break.

PR #435, opened 2025-10-08
Add Trigger.dev V4 template

jingerpie:
Tried the existed v3 template, was not able to get it to work, so I made the new template base of the official guide:
https://trigger.dev/docs/self-hosting/docker#setup

There is problems remain unresolved:
Problem 1: Domain/URL not accessible when deploying via template

When I deploy directly using my template, the domain URL fails to open
However, when I manually copy the same docker-compose.yml and .env files into a new compose service and manually configure the domain, it works perfectly
This suggests the template deployment process isn't handling the domain configuration correctly

Problem 2: PostgreSQL username configuration error

When I try to change the PostgreSQL username (from the default), I get this error:

  Fatal: role "postgresql" does not exist

It appears the system is still looking for a role named "postgresql" instead of using my custom username


Siumauricio:
I got error when importing in dokploy. 
image

eonminion:
@bbreunig The trigger V4 has problem when you deploying it for first time, but after you deploy and letting it run for sometime, and redeploy without any config changes, it will work. 

"Removing all "custom" networks and letting dokploy create the network via "isolated deployment" made trigger.dev v4 run smoothly."
This is likely after you remove it, you deploy once again, making it to work. I am not sure if this is dokploy issue or the compose script issue. Can you help to validate this?

eonminion:
@Siumauricio Can you kindly review this once again since it hasnt been merge for 2 months. pLease let me know what else I should fix?

bbreunig:
@eonminion I redeployed it multiple times without it getting better. Only after removing the extra networks did it finally work. A "race condition" with dokploy networks is much more likely than what you described. Coolify has the same issue. So far, trigger worked flawlessly for the past 2 weeks without the other networks so I don't see a reason to keep them in the docker-compose.

eonminion:
@bbreunig Thanks for getting back to me. To clarify, is your solution to remove this entire section? One of the reasons I initially added docker-proxy was because it allows access to dokploy’s database deployment on the same server without ever needing to expose the database. In my project, I deploy the app and database within the same project and allow trigger dev to accept jobs to interact with the database using an internal address. 

networks:
~~docker-proxy:~~
~~name: docker-proxy~~
  supervisor:
    name: supervisor
  webapp:
    name: webapp

oRafael07:
up

Siumauricio:
Close due to inactivity

LuizEdgar:
@jingerpie any chance to revive this?

PR #478, opened 2025-10-25
Fix documenso template

HowWhat42:
Fix and update documenso template to use the last version of the image.
Add more env variables in `docker-compose.yml` to fix errors on email verification.
Updated the `template.toml` to match the new env variables required.


HowWhat42:
@Siumauricio I can say by experience that using the current template caused me several errors and hours of searching/fixing.
First, SMTP errors on register and login, then S3 for API file upload and finally signing errors related to certificate.
You can check the documenso discord, I'm not the only one having problem with the template.

PR #481, opened 2025-10-27
Add Invoice Ninja

git-ari:
Adding Invoice Ninja template. You can redis to it but it is optional so I didn't want to add bloat to it.


Siumauricio (changes requested):
Check the env variables, probably your template.toml is empty

Screenshot 2025-11-07 at 11 44 21 PM

PR #523, opened 2025-11-10
Add Scrumlr io template

Shuna322:
Adds the Scrumlr web app template, allowing for running an online retrospective event.

Vibe coded with Github agent just for fun. 😳


Shuna322:
Hm can you double check if you followed instructions in the TOML for the private key generation? I didn't find a better place for those instructions 😅

PR #527, opened 2025-11-11
feat: Add anytype template

kipavy:


ByWaleed:
Just out of interest, I've tested this and it works as expected.

One step that may catch others is locating the client config for Anytype, which is stored inside the container at /data/client-config.yml. It would be helpful if this config could be output in the deployment log or—at minimum—a brief note added to the deployment description. For example, something like:

> After installation, run `cat /data/client-config.yml` in the service terminal to view the Anytype client config.

This is a small nitpick but it could save users some trouble. Appreciate the work and I hope you get Anytype published soon.

kipavy:
> Just out of interest, I've tested this and it works as expected.
> 
> One step that may catch others is locating the client config for Anytype, which is stored inside the container at /data/client-config.yml. It would be helpful if this config could be output in the deployment log or—at minimum—a brief note added to the deployment description. For example, something like:
> 
> > After installation, run `cat /data/client-config.yml` in the service terminal to view the Anytype client config.
> 
> This is a small nitpick but it could save users some trouble. Appreciate the work and I hope you get Anytype published soon.

It's not my project but thanks, about client-config.yml, yeah a Note in the description of the template can be a good idea, I'll see if I have the time to edit it but honestly its not that big of a deal considering anytype-bundle github repo README is pretty clear

blueprints/anytype/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+#
+# Usage:
+#   docker compose -f compose.aio.yml up -d
+#
+# The bundle image already contains MongoDB and Redis. Only the bundle service is required
+
+services:
+  any-sync-bundle:
+    image: ghcr.io/grishy/any-sync-bundle:latest
+    container_name: any-sync-bundle-aio
```
Siumauricio:
```suggestion
```

PR #550, opened 2025-11-25
feat: add Rote template

Rabithua:
- Add Rote deployment template with frontend, backend, and PostgreSQL services
- Configure domain routing for frontend (port 80) and backend (port 3000)
- Set up automatic password generation and environment variables
- Use latest image tag by default
- Add logo and metadata to meta.json

### DIFF
--- a/blueprints/rote/template.toml
+++ b/blueprints/rote/template.toml
@@ -18,6 +18,6 @@ host = "${backend_domain}"
 [config.env]
 POSTGRES_PASSWORD = "${postgres_password}"
 IMAGE_TAG = "${image_tag}"
-VITE_API_BASE = "https://${backend_domain}"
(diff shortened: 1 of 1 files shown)


blueprints/rote/template.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[[config.domains]]
+serviceName = "rote-backend"
+port = 3000
+host = "${backend_domain}"
+
+[config.env]
+POSTGRES_PASSWORD = "${postgres_password}"
+IMAGE_TAG = "${image_tag}"
+VITE_API_BASE = "https://${backend_domain}"
```
scanash00: Should be http:// as the traefik.me domain doesn't support https and it's the default domain set when deploying the template

Rabithua: done

PR #383, opened 2025-09-20
chore(blueprint): update Appwrite services to v1.8

poberbeck:
- Updating Appwrite to version 1.8
- Adding new environment variables
- Removing insecure default passwords from environment variables and replacing them with generated passwords

Sidenote: 
Unfortunately, we cannot set a wildcard domain for the new Appwrite Sites feature. You will need to add the subdomain manually for every site.

Closes #405

PR #551, opened 2025-11-25
Update Appwrite template to 1.8.0

adityaoberai:


jdlom:
Overall it works, but some routes are not correctly routed through Traefik. For example, when I create functions, the endpoint requests return a 404 instead of being forwarded to the right service.

It seems like certain paths aren't being picked up or matched by the Traefik rules defined in the template. Could you check if the routing configuration might be missing some rules or wildcard patterns for internal Appwrite services (like functions)?

poberbeck:
I also created a PR a few months ago (#383) and tested functions, sites, and the basic Ping/Pong example.
As far as I understand, Dokploy cannot acknowledge wildcard domains. For example, when using functions or sites that are deployed on random subdomains of the Appwrite service. 
After deploying a site or function, you need to add the specific domain manually in the “Domains” panel of Dokploy and reload the service.

PR #561, opened 2025-11-30
Dittofeed Template for dokploy

chrisspeakseh:
Dittofeed Template for dokploy
after trying different setups in CYBERSLIDE.net this is the best setup we found out to be working with all functions


Siumauricio:
Hey I get errors when trying to access to URL, also I see unhealthy checks can you check that? 
Screenshot 2026-02-15 at 1 22 37 AM

Siumauricio:
Close due to inacitivty

PR #576, opened 2025-12-08
add: restart policy to MinIO service

jemgdevp:
restart: unless-stopped is a Docker restart policy that automatically restarts a container if it stops due to an error or Docker daemon restart


russorat:
@malko @HoshangDEV you all made the most recent edits here. i know it's simple but any issues with this change?

PR #578, opened 2025-12-11
Refactor ‎blueprints/bigcapital/docker-compose.yml for service updates

adryserage:
Updated docker-compose.yml to modify service configurations, including changing ports to expose and adding a new database migration service.


Siumauricio:
Close due to inactivity

PR #579, opened 2025-12-12
feat(blueprint): peerdb template

nktnet1:
Made based on:
- https://github.com/PeerDB-io/peerdb/blob/main/docker-compose.yml


nktnet1:
Connecting to `dokploy-network` was necessary for me to see other containers also deployed in Dokploy.

PR #582, opened 2025-12-12
Fix Plane env URLs

adryserage:
## Summary
- set Plane WEB_URL to include https scheme
- point API_BASE_URL to the public proxy endpoint using https

## Testing
- not run (not needed for env template changes)

### DIFF
--- a/blueprints/plane/template.toml
+++ b/blueprints/plane/template.toml
@@ -9,7 +9,7 @@ secret_key = "${base64:48}"
 [config]
 env = [
 "Domain=${domain}",
-"WEB_URL=${Domain}",
+"WEB_URL=https://${Domain}",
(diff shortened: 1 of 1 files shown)


blueprints/plane/template.toml
```diff
@@ -9,7 +9,7 @@ secret_key = "${base64:48}"
 [config]
 env = [
 "Domain=${domain}",
-"WEB_URL=${Domain}",
+"WEB_URL=https://${Domain}",
```
Siumauricio, line 12: why https? by default all the traefik domains are http

Siumauricio (changes requested):

Siumauricio:
Close due to inactivity

PR #595, opened 2025-12-16
fix: superset template fixed versions, postgres volume mount and network instructions

nktnet1:
## What is this PR about?

Updating postgres template - fixed version.

Also fixed postgres volume mount which changed in v18, and simplified network instructions.

## Checklist

### DIFF
--- a/blueprints/superset/docker-compose.yml
+++ b/blueprints/superset/docker-compose.yml
@@ -15,23 +15,14 @@
 # ## NETWORK INSTRUCTIONS
 #
 # If you want to connect superset with other internal databases managed by
-# Dokploy (on dokploy-network) using internal hostnames, you will need to
-# uncomment the `networks` section, both for the superset container and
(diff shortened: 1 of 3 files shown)


Siumauricio:
For some reason is throwing internal server error 

Screenshot 2026-02-15 at 7 13 28 PM
Screenshot 2026-02-15 at 7 13 53 PM

nktnet1:
@Siumauricio did you set up with either

```
superset-init
```

or

```
superset-demo
```

inside the container after deploying?

A database migration is necessary, even if you've already initialised and are only updating, since it's a major version change.

I've tried deploying using this template just now (copied the preview URL as the template base url in Dokploy) and it works:

image

&nbsp;

You may also want to test on `HTTPS` instead of `HTTP`, and clear existing superset volume/config.

Siumauricio:
That works, thank you @nktnet1

PR #664, opened 2026-01-18
chore: updating and cleaning up several templates.

rubixvi:
Updating and house keeping on a number of templates.


Siumauricio:
Hey, I appreciate the PR, but I'm not a fan of testing 23 templates. It would take me a long time, since I see that several versions are changing at once. This could substantially break some templates. If you could create a PR for each template, that would be ideal.

I will close this for now, thank you

PR #770, opened 2026-04-04
[codex] Improve Crawl4AI self-hosting template

Manan-Santoki:
## What changed

- pin the Crawl4AI image to `0.8.0`
- add OpenRouter-ready OpenAI-compatible env settings
- mount a generated `/app/config.yml` with documented production-oriented defaults
- enable JWT auth, rate limiting, health/metrics observability, and stricter healthchecks
- update the template metadata and docs link to match the current self-hosting docs

## Why

The existing blueprint was still on `latest`/`0.7.3` and only exposed basic provider env vars. It did not reflect Crawl4AI's current `0.8.0` self-hosting guidance around pinned versions, OpenAI-compatible base URLs, mounted `config.yml`, or built-in security and observability features.

## Impact

- Dokploy users can point Crawl4AI at OpenRouter by setting `OPENAI_API_KEY` to an OpenRouter key and using the default `OPENAI_BASE_URL`
- the deployment now ships with JWT auth enabled by default instead of being exposed without app-layer auth
- the template better matches the current Crawl4AI self-hosting docs and feature set

## Validation

- ran `node dedupe-and-sort-meta.js`


Siumauricio:
Closing this draft as part of a cleanup of stale pull requests — it's been inactive for several months. If you're still interested, feel free to update and reopen against the latest canary. Thanks for the contribution!

PR #865, opened 2026-05-15
feat: add powersync blueprint template

ahmed-lotfy-dev:
Adds a PowerSync blueprint template for Dokploy.

PowerSync is an open-source sync engine for local-first apps. It syncs PostgreSQL data to mobile devices and browsers with real-time replication.

**Template features:**
- Single service (no MongoDB needed) — uses PostgreSQL for both source and bucket storage
- All configuration through environment variables (POWERSYNC_CONFIG_B64 / POWERSYNC_SYNC_CONFIG_B64)
- Pinned to journeyapps/powersync-service:1.21.0
- Service name matches blueprint folder name (powersync)

**Checklist:**
- [x] Logo is valid PNG
- [x] Docker image pinned to specific version
- [x] Service name matches blueprint folder
- [x] meta.json updated with correct format
- [x] No accidental deletions from meta.json


Siumauricio:
**Automated template check** 🤖

This PR's template was imported into a test Dokploy instance (using the `docker-compose.yml` + `template.toml` from the PR head) and deployed, and its configured domains were then checked over HTTP.

**Result: ❌ some issues need attention**

### `powersync`
- Deploy: `done` (37s)
- Domain checks:
  - `powersync:8080` → HTTP 404 ❌
- Container states at failure time: powersync-1: restarting
- Issues:
  - domain check failed for service `powersync:8080` (path `/`): kept returning HTTP 404 for 3 minutes after the deploy finished

_The test service is removed automatically after verification. Happy to re-test once the PR is updated._

Siumauricio:
I tested this template on a Dokploy instance and it failed to deploy: the `powersync` container crash-looped and the domain kept returning 404. Root cause: `PS_DATA_SOURCE_URI` defaulted to a placeholder connection string (`postgresql://user:password@host:5432/db`), so the service could never connect to a database and died on startup.

I pushed a fix to this branch (84801304):

- **Bundled Postgres 17** with `wal_level=logical`, a healthcheck, and a persistent volume, so the template works out of the box. `powersync` waits for it to be healthy before starting.
- **Readable `powersync.yaml` via `[[config.mounts]]`** instead of the base64-encoded env var. Replication reads from the `powersync` database; bucket storage uses a separate `powersync_storage` database (PowerSync's Postgres storage must not share a database with the replication source).
- **Init SQL mount** that creates the storage database, a demo `lists` table synced by the default sync rules, and the `powersync` publication required for logical replication.
- **Replaced the `sync_config` block with standard `sync_rules`** (the previous key isn't valid config for `powersync-service` 1.21.0) and added a generated HS256 dev key so client auth works out of the box.
- **Removed the host port mapping** (Traefik routes to the container port directly) and pointed the template domain at `/probes/liveness`, since PowerSync serves no route on `/`.

Verified on Dokploy: deploy completes, replication of the demo table starts, and the domain returns HTTP 200 on `/probes/liveness`.

PR #880, opened 2026-05-19
Update forgejo template; bumps Forgejo version to 15; fixes git push over ssh

DiegoEnriquezSerrano:
## What is this PR about?

* Bumps forgejo version from `10` -> `15` (current major)
* Fixes ssh conflict with host default ssh listening port.
  * Current configuration prevents `git push` over ssh from working due to invalid rw privileges. Assuming host ssh server is listening on default port `:22`, DNS resolution to dokploy service domain will resolve to host IP address, ssh request will be routed to host ssh server instead of container ssh server.

## Checklist

- [ x] I have read the suggestions in the README.md file https://github.com/Dokploy/templates?tab=readme-ov-file#general-requirements-when-creating-a-template
- [ ] I have tested the template in my instance, so the maintainers don't spend time trying to figure out what's wrong.


Siumauricio:
Thanks for the PR and sorry for the long wait! I deploy-tested this on a live Dokploy instance and pushed a follow-up commit (d600245) to get it working end to end. Summary:

**Tested as submitted** (Forgejo 15): the web UI deployed fine (HTTP 200, installer page), but the second `[[config.domains]]` entry on port 22 returned `500 Internal Server Error`. Dokploy domains are Traefik **HTTP** routers, so they can't proxy SSH traffic — that domain can never work, unfortunately. Exposing SSH on a host port is something users need to set up themselves for their instance (same approach as our Gitea templates, which just `expose` port 22).

**What I changed in the follow-up commit** (re-tested, deploys green with both containers running and the installer correctly pre-filled with the assigned domain thanks to your `FORGEJO__server__DOMAIN`/`ROOT_URL` additions):

- Bumped `codeberg.org/forgejo/forgejo:15` → `:16` — Forgejo moved on while this sat in draft; 16.0.2 is the current stable.
- Removed the port-22 domain from `template.toml` and the comment/blank entries in the `env` array.
- `ports:` → `expose:` — since #1053 the repo convention is to not publish host ports; Traefik reaches containers over the docker network.
- Dropped the empty `FORGEJO__server__SSH_PORT=` and `SSH_LISTEN_PORT=22` overrides (those are already the image defaults, and an empty value is risky).
- Kept your version bump, `depends_on` fix, and the `DOMAIN`/`SSH_DOMAIN`/`ROOT_URL` environment wiring — that part is a real improvement over the current template.

If this looks good to you, please mark the PR as ready for review and we can merge it.

Issue #152, opened 2025-05-25
Get Your Custom Dokploy Templates Built!


Siumauricio:
## Hey everyone,

We've got some exciting news about how we're speeding up the creation and integration of open-source templates for Dokploy. We're officially using [Algora](https://algora.io/Dokploy/dashboard) to manage bounties, and we want you to be a part of it!

### Need a Template? Get It Built!

If you have a need for a template e.g. invoiceninja, wordpress etc, and you don't have the necessary knowledge but need a template and want someone to do the implementation in dokploy, you can use bounties using  [Algora](https://algora.io/Dokploy/dashboard)

### How to Create a Bounty:

just typing `/bounty $1000` in an issue comment will automatically create a bounty for the issue,

### For Bounty Creators:

To ensure the best results, please make sure to:

- **Provide clear specifications** about what you expect from the template.
- **Include any relevant documentation**—this will be a huge help to potential contributors!

This initiative will help us build a wider variety of templates, making Dokploy even more powerful and easy to use. Plus, it provides a great motivation for contributors – **free money**! 🤑

debaa98:
hey @Siumauricio is this bounty still open ?

Artemis-ignis:
Submitted PR #881 for this bounty: https://github.com/Dokploy/templates/pull/881

It adds a Miniflux template with pinned images, PostgreSQL persistence, Dokploy domain/env config, SVG logo, and a sorted meta.json entry.

Validation currently posted in the PR:
- `node dedupe-and-sort-meta.js` passes with 390 entries and 0 duplicates.
- `npm run validate-template -- --dir ../blueprints/miniflux` passes.
- `npm run validate-docker-compose -- --file ../blueprints/miniflux/docker-compose.yml` passes.
- `pnpm exec tsc -b` passes for the preview app.
- Cloudflare preview is Ready for commit `dbd929d`.
- Chrome preview check shows the Miniflux card and no console errors.

I have not tested importing it into a live Dokploy instance yet, but I can follow up quickly if that is required.

zhaog100:
🙋 Claiming this bounty. I'll start work within 24-48h.

NightVibes33:
/attempt #152

Opened PR #883: https://github.com/Dokploy/templates/pull/883

Current scope: 11 Dokploy templates, each with metadata, icon, domain config, and Compose files following repo validators.

BinkyTwin:
/attempt #152

[...]

PR #898, opened 2026-05-21
feat: add code-server template

MANFIT7:
@algora-pbc /claim #152

Part of #152
Related #145

## Summary
- adds a code-server blueprint using `codercom/code-server:4.121.0`
- wires Dokploy domain/env variables for browser access, generated password, timezone, and persistent home/config volumes
- adds the code-server icon and `meta.json` entry with docs, website, and GitHub links

## Validation
- `node dedupe-and-sort-meta.js`
- `PATH="/Applications/Codex.app/Contents/Resources:$PATH" npm run validate-template -- --dir ../blueprints/code-server`
- `PATH="/Applications/Codex.app/Contents/Resources:$PATH" npm run validate-docker-compose -- --file ../blueprints/code-server/docker-compose.yml`
- `git diff --check`

Note: Docker is not installed in my local environment, so I could not run `docker compose config` locally. The repository template and compose validators pass.


Siumauricio:
**Automated template check** 🤖

This PR's template was imported into a test Dokploy instance (using the `docker-compose.yml` + `template.toml` from the PR head) and deployed, and its configured domains were then checked over HTTP.

**Result: ❌ some issues need attention**

### `code-server`
- Deploy: `done` (24s)
- Domain checks:
  - `code-server:8080` → HTTP 404 ❌
- Container states at failure time: code-server-1: restarting
- Issues:
  - domain check failed for service `code-server:8080` (path `/`): kept returning HTTP 404 for 3 minutes after the deploy finished

_The test service is removed automatically after verification. Happy to re-test once the PR is updated._

Siumauricio:
Pushed a fix for the crash loop that was causing the sustained Traefik 404.

**Root cause:** the named volume mounted at `/home/coder/.config` was created root-owned (that path does not exist in the `codercom/code-server` image), so code-server, running as uid 1000, crashed on startup with:

```
error EACCES: permission denied, mkdir '/home/coder/.config/code-server'
```

The container restarted in a loop, so the Traefik router never had a healthy backend.

**Fix:** removed the nested `code-server-config` volume. `.config` already lives inside `/home/coder`, which is persisted by the existing `code-server-home` volume, so nothing is lost.

**Verified on a Dokploy instance:** deploy completes, the container stays up (`HTTP server listening on http://0.0.0.0:8080/`, authentication enabled via `$PASSWORD`), and the generated domain returns HTTP 200 with the code-server login page.

PR #960, opened 2026-06-25
feat: add Hermes (Nous Research) agent template

vikyw89:
## Hermes

Adds a template for [Hermes](https://hermes-agent.nousresearch.com/), the open agent runtime from Nous Research — an OpenAI-compatible gateway API plus a supervised web dashboard.

Based on the upstream [Docker Compose example](https://hermes-agent.nousresearch.com/docs/user-guide/docker), adapted to Dokploy conventions.

### What it exposes
- **Dashboard** (port 9119) — primary UI; configure provider keys, skills, and memories here.
- **Gateway API** (port 8642) — OpenAI-compatible endpoint.

### Notable, doc-grounded choices
- The gateway API is **disabled and loopback-bound by default** upstream, so it would be unreachable behind Traefik. The template sets `API_SERVER_ENABLED=true`, `API_SERVER_HOST=0.0.0.0`, and a generated `API_SERVER_KEY` (bearer token) so the published `8642` domain actually works.
- Hermes' dashboard auto-enters **auth-gated mode** once it binds off-loopback (which it does behind Traefik). The template wires up the bundled basic-auth provider (`HERMES_DASHBOARD_BASIC_AUTH_USERNAME=admin` + a generated 128-char password and session secret) so the public domain isn't left unauthenticated.
- Image pinned to `nousresearch/hermes-agent:v2026.6.19`; `meta.json` version matches.
- Config/keys persist in the `hermes_data` volume.

### ⚠️ Security — enable HTTPS before use
Dokploy creates template domains over plaintext HTTP by default (`certificateType: "none"`). Until you enable TLS, the dashboard credentials **and** the gateway bearer token travel in cleartext. After deploy, edit each domain in **Dokploy → Domains** to enable **HTTPS + Let's Encrypt** (with HTTP→HTTPS redirect). The `*.sslip.io` host resolves to your server IP, so the certificate issues automatically. Only then is the basic-auth / bearer-token protection meaningful.

### After deploy
Open the dashboard domain (user `admin`, password from the generated env var) and configure a model provider key (e.g. Anthropic/OpenAI). The API bearer token is the generated `API_SERVER_KEY` env var.

Validated locally with the repo's `validate-template.ts`, `validate-docker-compose.ts`, and `process-meta.js` — all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mrmuminov:
+1 for this. I'd really appreciate having an official Hermes Agent template in Dokploy.

PR #990, opened 2026-07-12
fix(billmora): add shared volumes for plugins and themes across containers

Nekoplii:
## What is this PR about?
This PR updates the Billmora template to mount shared named volumes for `plugins `and `themes`. This is required to synchronize dynamically installed user assets across the isolated app, worker, and scheduler containers.

New PR of Billmora

## Checklist

### DIFF
--- a/blueprints/billmora/docker-compose.yml
+++ b/blueprints/billmora/docker-compose.yml
@@ -28,6 +28,9 @@ services:
       - AUTORUN_LARAVEL_STORAGE_LINK=true
     volumes:
       - billmora-storage:/var/www/html/storage
+      - billmora-plugins:/var/www/html/plugin
+      - billmora-themes-resources:/var/www/html/resources/themes
+      - billmora-themes-public:/var/www/html/public/themes
     depends_on:
       billmora-db:
         condition: service_healthy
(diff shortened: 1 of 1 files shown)


Siumauricio:
Tested on a Dokploy instance — works. ✅

**Deploy evidence** (template imported and deployed as-is from this branch):
- Deploy: `done` (77s) · all 5 containers running (app, worker, scheduler, db, redis)
- Root URL → HTTP 200, `<title>Billmora</title>`
- Theme asset served from the shared volume: `GET /themes/portal/moraine/css/style.css` → HTTP 200

**Volume paths verified against `ghcr.io/billmora/billmora:latest`:**
- `/var/www/html/plugin` → exists, `www-data:www-data` (Gateways/Modules/Provisionings/Registrars)
- `/var/www/html/resources/themes` → exists (admin/client/email/invoice/portal)
- `/var/www/html/public/themes` → exists (built Vite assets: admin/client/portal)

All three paths ship with content in the image, so Docker's copy-on-first-use populates the named volumes and the app boots normally. The app installs plugins/themes at runtime into exactly these paths (`base_path("plugin/...")`, `resource_path("themes/...")`, `public_path("themes/...")`), so sharing them across app/worker/scheduler is the right fix for keeping user-installed assets in sync.

Minor heads-up (not a blocker): since these directories are now persisted in volumes, bundled themes/plugins from a newer image will be masked by the existing volume contents after upgrades — updates to those assets need to go through the app. Standard trade-off for this pattern.

PR #1043, opened 2026-07-23
Add Creed self-host template

Simciutoja:
[Creed](https://creed.md/)  is your personal context profile. One file that captures who you are: values, goals, work, preferences, constraints, people, health, routines. Any AI you talk to knows you instantly instead of starting from zero every conversation

Summary
- Adds a new `creed` template (`blueprints/creed/`) for self-hosting [Creed](https://github.com/connorhpbrn/creed) - a personal context file (creed.md) that AI agents read before answering and propose updates to.
- Builds the app from source via an inline multi-stage Node 20 Dockerfile (the upstream repo doesn't ship an official image), since it's a stateless Next.js app.
- Only `CREED_ENCRYPTION_SECRET` is auto-generated. Creed requires a real Supabase project (Auth, RLS, realtime, schema pushed via `supabase db push`), which can't be provisioned from within a docker-compose deploy - so Supabase/Stripe/OpenRouter/Resend/GitHub OAuth vars are left blank, with step-by-step setup instructions in the mounted `README.md`.
- `NEXT_PUBLIC_*` vars are passed as Docker build args (not just container env), since Next.js inlines them at build time and the app fails the build outright if `NEXT_PUBLIC_SITE_URL` is unset.

## Test plan
- [x] `node build-scripts/generate-meta.js --check` passes (501 templates validated)
- [x] `docker compose build --no-cache` succeeds against the upstream repo
- [x] `docker compose up` boots the container; `GET /` returns 200
- [x] `GET /api/health` returns 200 with `status: "degraded"` and `db.error: "supabase_admin_not_configured"` when Supabase is unset, matching the documented behavior
- [ ] Preview-link BASE64 import test in a live Dokploy instance (per repo README step 9)


Siumauricio:
Thanks for the contribution — the template is well put together (the mounted README with the Supabase setup steps is a nice touch), but as-is it fails to deploy on a live Dokploy instance for two independent reasons. I tested it today and both are confirmed, along with the fixes:

**1. The build context must stay pinned.** In 5b4c211 you reverted the pin from 1b8b9b3 back to `main`, and upstream moved to a monorepo on 2026-08-02 (`5260f7f "move creed into a monorepo"`), which now breaks the inline Dockerfile:

```
0.659 sh: tsc: not found
0.661 npm error Lifecycle script `build` failed with error:
0.661 npm error code 127
0.661 npm error workspace creed-cli@0.2.2
failed to solve: process "/bin/sh -c npm run build" did not complete successfully: exit code: 127
```

Even if the root build passed, the runner stage would still break at HEAD: the Next.js output now lives in `apps/creed/.next` (not `/app/.next`), there is no root `next.config.ts`, and the root `package.json` has no `start` script. Please restore the pin (it also matches the `"version": "e9b096d"` you already declare in `meta.json`; version bumps can update the pin deliberately later):

```yaml
      context: https://github.com/connorhpbrn/creed.git#e9b096d6d68fef6df336338bc9965a94ab62f75e
```

**2. The `env` array in `template.toml` is in the wrong TOML table.** It's placed after the `[[config.domains]]` header, so it parses as a key *inside* the domain entry instead of `config.env`. Dokploy imports the template with an empty environment — every `${VAR}` in the compose resolves to an empty string, and even with the pin restored the build dies with:

```
[cause]: Error: NEXT_PUBLIC_SITE_URL is not set. Set it to the deployed origin (e.g. https://your-deploy.vercel.app) ...
> Build error occurred
Error: Failed to collect page data for /_not-found
```

Fix: move the `env = [ ... ]` block up so it sits directly under `[config]`, before `[[config.domains]]` (see e.g. `blueprints/activepieces/template.toml` for the convention).

With exactly those two changes I deployed the template on a live Dokploy instance and it works end to end: the build succeeds, `GET /` returns 200 with the Creed landing page, and `GET /api/health` returns the documented degraded state:

```json
{"status":"degraded","components":{"api":{"ok":true},"db":{"ok":false,"error":"supabase_admin_not_configured"},"auth":{"ok":false,"error":"supabase_admin_not_configured"}}}
```

Everything else checks out: `generate-meta.js --check` passes, no host ports / `container_name` / external networks, and the README mount imports cleanly. Could you push those two changes? Happy to merge after that.

Simciutoja:
Thanks for the detailed review and for verifying the deploy, it's my first time contributing to so big projects, so is my first mistake :>

Both fixes are in:
1. Restored the build context pin to `e9b096d` (matches `meta.json` version)
2. Moved `env` under `[config]`, before `[[config.domains]]`

This is my first time contributing a template like this, so sorry for the mistakes - thanks again for catching them and walking through the fixes. Happy to adjust further if anything else comes up.

Siumauricio:
Re-tested at ba0a906 after the fixes — both issues are resolved:

- Build context is now pinned to commit `e9b096d`, matching `meta.json`'s version.
- `env` now sits under `[config]` (parses as `config.env`), no longer inside `[[config.domains]]`.

Deployed on a test instance: build from source completed in ~3 min, container running, `GET /` returns 200 («Creed - Personal context for your agents.») and `GET /api/health` returns 200 with `supabase_admin_not_configured` for db/auth, which is the documented behavior until an external Supabase project is configured. LGTM.

PR #1057, opened 2026-07-29
Update SendDock template (track latest, drop deprecated env vars)

JuansesDev:
Keeps the SendDock template current and cleans up two variables deprecated in 0.8.

**Changes**
- Point the image at `ghcr.io/arkhe-systems/senddock:latest` so fresh deploys always get the current release — no per-release PR needed to keep the gallery up to date.
- Remove `DEPLOYMENT_MODE` — self-host is the default now (the hosted flag became the boolean `CLOUD`).
- Remove `SENDDOCK_LICENSE_KEY` (from `docker-compose.yml` and `template.toml`) — the Pro license is activated from the dashboard (Instance settings) in 0.8; the env var is removed in 0.9.
- `meta.json` version set to 0.8.0 as a reference baseline.

No changes to services, ports, domains or volumes. `node build-scripts/generate-meta.js --check` passes (500 templates validated).

### DIFF
--- a/blueprints/senddock/docker-compose.yml
+++ b/blueprints/senddock/docker-compose.yml
@@ -27,7 +27,7 @@ services:
       retries: 5
 
   senddock:
-    image: ghcr.io/arkhe-systems/senddock:0.6.5.1
+    image: ghcr.io/arkhe-systems/senddock:latest
     restart: unless-stopped
     depends_on:
       postgres:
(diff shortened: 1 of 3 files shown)


Siumauricio:
Tested this on a live Dokploy instance — deploys and runs correctly:

- Deploy finished in ~21s, all 3 services healthy (postgres, redis, senddock).
- `GET /` → HTTP 200, serves the SendDock UI (`<title>SendDock</title>`).
- `GET /health` → HTTP 200 `{"status":"ok"}`, DB check passing — confirms the app boots fine without `DEPLOYMENT_MODE` and `SENDDOCK_LICENSE_KEY`.
- Verified on ghcr that `latest` currently resolves to the same digest as `0.8.0` (`sha256:1450a162…`), matching the `meta.json` baseline.

One note (non-blocking): the repo generally prefers pinned image tags over `latest` for reproducible deploys, though plenty of templates do track `latest`. Since ghcr also publishes `0.8` and `0` tags, pinning `0.8.0` (or tracking `0`) would be a middle ground if maintainers prefer — fine either way from the testing side.

PR #1079, opened 2026-08-12
feat: add DockRoute template

danilodorgam:
Adds [DockRoute](https://www.dockroute.dev) ([source](https://github.com/Dockroute/Dockroute), MIT) — External-DNS for plain Docker hosts. It watches running containers, reads `dockroute.*` labels and reconciles the matching DNS records (and Cloudflare Tunnel routes) in a pluggable provider, with ExternalDNS-style TXT ownership so it never alters records it cannot prove it manages.

Positioning for Dokploy users: DNS automation for the containers your platform doesn't route — internal/LAN DNS, Cloudflare Tunnel ingress, or records for services outside the Traefik flow.

Template notes:
- Headless service (no UI), so `domains = []` — same shape as `cloudflare-ddns` / `cloudflared`.
- Defaults are safe: the `log` provider is a zero-credential dry run that only prints the desired state; switching `DOCKROUTE_PROVIDER` to `cloudflare` plus an API token goes live.
- Docker socket is mounted read-only; the image detects the socket's group at runtime and drops privileges to a non-root user by itself.
- Image `ghcr.io/dockroute/dockroute:0.2.0`, multi-arch (amd64 + arm64), Trivy-gated releases.

Validated with `build-scripts`: `validate-template` and `validate-docker-compose` both pass.


Siumauricio:
Reviewed and tested this template on a live Dokploy instance.

**Legitimacy check**
- Upstream repo [Dockroute/Dockroute](https://github.com/Dockroute/Dockroute) is young (created 2026-07-28) but real: public TypeScript source with tests, CI/CodeQL, semantic-release, 3 tagged releases, MIT license.
- Image `ghcr.io/dockroute/dockroute:0.2.0` is publicly pullable, multi-arch, and its OCI `revision` label (`7bc0197`) matches an actual commit in the repo.
- Author account dates from 2013 with a consistent history.

**Docker socket note**
The template mounts `/var/run/docker.sock` read-only, which is inherent to what the tool does (watching container labels, external-dns style). This is in line with existing templates (portainer, dockge, dozzle, watchtower-class tools — several of which mount the socket read-write). The entrypoint drops root to a non-root user after picking up the socket group.

**Deploy test (Dokploy instance)**
- Template imported and deployed cleanly, container `running` with `RestartCount: 0`, no restart loop. Headless service, `domains = []` as expected (same shape as `cloudflare-ddns` / `cloudflared`).
- Found one functional bug: the compose file had no `environment:` block, so none of the `DOCKROUTE_*` / `CLOUDFLARE_*` variables from `template.toml` reached the container (verified via container inspect — Env only had PATH/BUN vars). Setting `DOCKROUTE_PROVIDER=cloudflare` plus a token in the Dokploy environment tab would have silently stayed in the log dry-run mode.
- Pushed a fix to this branch mapping the template variables through `environment:`. Redeployed with the fix: all variables now present in the container (`DOCKROUTE_PROVIDER=log`, `DOCKROUTE_POLICY=sync`, etc.), container stable.

`generate-meta.js --check` passes (515 templates). LGTM after the pushed fix.

PR #1080, opened 2026-08-12
Add Didacta template

va360labs:
## What is this PR about?

New PR of **Didacta** — a fair-code, self-hosted LMS: course authoring, a public catalog with checkout, enrollment and invitation flows, gamification, certificates, reporting and a module marketplace. Tenants are isolated at the database level with PostgreSQL RLS.

The blueprint ships the app image plus Postgres and Redis, and handles the two things that are easy to get wrong when wiring this by hand:

- **Postgres must be `pgvector/pgvector:pg16`.** The baseline migration runs `CREATE EXTENSION "vector"`, so a plain `postgres:16` fails on first boot with no obvious clue.
- **One domain is enough.** The container runs the API on `:4000` and the web on `:3000`, and Next.js rewrites `/api/*`, `/healthz` and `/readyz` to the internal API, so `[[config.domains]]` only targets port 3000.

`DIDACTA_SETUP_TOKEN` is generated with `${password:32}`. It protects the creation of the first admin account while the instance is still empty; generating it here means the user opens `https://<domain>/setup?token=<DIDACTA_SETUP_TOKEN>` instead of digging it out of the container logs.

`TENANT_SETTINGS_ENC_KEY` is deliberately left unset: the app generates a 32-byte hex key on first boot and persists it in the data volume. Setting it from the blueprint with a helper that does not produce exactly 64 hex characters would break secret encryption (SSO, per-tenant SMTP, Stripe) silently.

Links: [didacta.io](https://didacta.io) · [docs](https://docs.didacta.io) · [source](https://github.com/va360labs/didacta-io) · image `ghcr.io/va360labs/didacta-community` (public, no login required).

## Checklist

- [x] I have tested the template in my instance, so the maintainers don't spend time trying to figure out what's wrong. — see **Testing** below for exactly what was exercised. I don't run a Dokploy instance, so I tested this exact `docker-compose.yml` with the `.env` that `template.toml` generates, and deployed the same stack on another PaaS end to end. The PR preview is the last mile.
- [ ] I have added tests that demonstrate that my correction works or that my new feature works. — not applicable to a template.

`node build-scripts/generate-meta.js --check` passes (515 templates validated).

None.

## Testing

Brought up this exact `docker-compose.yml` with the `.env` produced from `[config.env]` (resolving `${domain}`, `${password:64}` and `${password:32}` the way Dokploy does), on Docker Compose v2:

- First boot applied migrations, enabled pgvector, applied the RLS policies and the role grants, and logged that the runtime connection is derived to the non-superuser role — which is what makes tenant isolation real.
- `GET /healthz` and `GET /api/v1/setup/status` both answered **through port 3000**, confirming a single domain is all the proxy needs.
- Creating the first tenant and admin account via `POST /setup/init` succeeded, and the public catalog and sign-in pages rendered.

The same image and topology were then deployed on a self-hosted Coolify instance (Docker Compose), where all three containers reached healthy and the first-run setup link worked over a real domain — so the app side of this blueprint is verified on a proxy-fronted PaaS, not just locally.

## Screenshots or Videos

The course editor of the running instance:

Didacta running


Siumauricio:
Tested this template end-to-end on a real Dokploy instance. Everything works out of the box.

**Deployment evidence**
- Template imported and deployed via the Dokploy API (compose + template.toml as-is from this PR).
- All 3 containers reached healthy: `didacta` (Up, healthy), `postgres` (pgvector/pgvector:pg16, healthy), `redis` (healthy).
- HTTP through the generated domain (Traefik → didacta:3000):
  - `GET /` → `307` → `/setup` (uninitialized instance, expected)
  - `GET /setup` → `200`
  - `GET /healthz` → `200` `{"status":"ok","service":"api","version":"0.0.1-alpha.107",...}` — confirms the Next.js rewrite to the internal API works through a single domain
  - `GET /api/v1/setup/status` → `200` `{"initialized":false}`

**Review notes**
- Compose follows repo conventions: no `ports:`, no `container_name:`, no custom networks, pinned image tag (`ghcr.io/va360labs/didacta-community:0.0.1-alpha.107`, public and pullable), `restart: unless-stopped` everywhere.
- `template.toml` domain targets the right service/port (`didacta:3000`), secrets generated with `${password:N}` helpers, and `DIDACTA_CORE_VERSION` matches the pinned image tag.
- `node generate-meta.js --check` passes (515 templates validated).
- Verified the image corresponds to the public source repo (OCI labels point to `va360labs/didacta-io`, revision-pinned) and the project has a real development history (image tags back to `0.0.1-alpha.51`, built 2026-05-06).
- Nice touch documenting why `pgvector/pgvector:pg16` is required and why `TENANT_SETTINGS_ENC_KEY` is intentionally left unset.

LGTM.

PR #1081, opened 2026-08-13
feat: add Louez template

loan-talvat:
## Louez

[Louez](https://github.com/Synapsr/Louez) is an open-source equipment rental platform (AGPL-3.0). Rent out cameras, tools, bikes, party gear or vehicles from a branded storefront with live availability, and manage products, reservations, pickups and returns, customers, PDF contracts and revenue statistics from one dashboard. UI in 8 languages.

## What the blueprint deploys

| Service | Image | Role |
|---|---|---|
| `louez` | `synapsr/louez:2.1.0` | Storefront, dashboard and API — installs its own schema on first boot |
| `mysql` | `mysql:8.4` | Database |
| `minio` | `minio/minio:RELEASE.2025-09-07T16-13-09Z` | Product images, private bucket streamed same-origin through the app under `/files` |

Generated credentials come from `${password:32}` helpers, the domain from `${domain}`. The app's `AUTH_SECRET` is generated on first boot and persisted in the `louez-data` volume, so sessions survive redeploys.

## Testing

Deployed and exercised locally with this exact `docker-compose.yml` before opening the PR:

- three services healthy, database schema installed automatically on first boot;
- first-run routing verified (`/` lands on the login flow, the store resolver 404s on an empty instance);
- account created through the UI flow, product image uploaded through the app into the MinIO bucket, and served back byte-identical from `/files`;
- `node build-scripts/generate-meta.js --check` → `✅ 515 templates validated`.

One note for reviewers, in case it helps other MinIO-based blueprints: the bucket is created by the storage service's start command rather than a one-shot `mc` job, and that requires replacing the image entrypoint (`entrypoint: ["/bin/sh", "-c"]`). MinIO's own entrypoint prepends `minio` to whatever arguments it receives, so a plain `command: sh -c '…'` silently becomes `minio sh -c '…'` and the container crash-loops.

### DIFF
--- a/blueprints/louez/docker-compose.yml
+++ b/blueprints/louez/docker-compose.yml
@@ -1,0 +1,71 @@
+version: "3.8"
+
+services:
+  louez:
+    image: synapsr/louez:2.1.0
+    restart: unless-stopped
+    environment:
+      LOUEZ_MODE: standalone
+      DATABASE_URL: mysql://louez:${DB_PASSWORD}@mysql:3306/louez
+      AUTH_URL: https://${LOUEZ_HOST}
+      AUTH_TRUST_HOST: "true"
+      NEXT_PUBLIC_APP_URL: https://${LOUEZ_HOST}
+      NEXT_PUBLIC_APP_DOMAIN: ${LOUEZ_HOST}
+      # The bucket stays private: Louez streams images same-origin under /files
+      # with its own credentials, so the store needs no domain of its own.
+      S3_ENDPOINT: http://minio:9000
+      S3_REGION: us-east-1
+      S3_BUCKET: louez
+      S3_ACCESS_KEY_ID: louez
+      S3_SECRET_ACCESS_KEY: ${MINIO_PASSWORD}
+      S3_PUBLIC_URL: /files
+    volumes:
+      # Holds the auth secret generated on first boot, so sessions survive
+      # redeploys.
+      - louez-data:/app/data
+    depends_on:
+      mysql:
+        condition: service_healthy
+      minio:
+        condition: service_healthy
+
+  mysql:
+    image: mysql:8.4
+    restart: unless-stopped
+    environment:
+      MYSQL_DATABASE: louez
+      MYSQL_USER: louez
+      MYSQL_PASSWORD: ${DB_PASSWORD}
+      MYSQL_ROOT_PASSWORD: ${DB_ROOT_PASSWORD}
+    volumes:
+      - mysql-data:/var/lib/mysql
+    healthcheck:
+      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
+      interval: 10s
+      timeout: 5s
+      retries: 10
+
+  minio:
+    image: minio/minio:RELEASE.2025-09-07T16-13-09Z
+    restart: unless-stopped
+    # A bucket is a top-level directory of the data drive; creating it here
+    # keeps the stack to three services instead of adding a one-shot job. The
+    # entrypoint is replaced because the image's own prepends `minio` to its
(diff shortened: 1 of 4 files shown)


loan-talvat:
Thanks for the automated review — I checked both findings against Dokploy's own source, and I believe both are false positives. Evidence below so a maintainer doesn't have to re-derive it.

**1. "Generate a URL-safe database password"**

`${password:N}` cannot produce URI-reserved characters. From `Dokploy/dokploy`, `packages/server/src/templates/index.ts`:

```ts
export const generatePassword = (quantity = 16): string => {
	const characters =
		"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
	let password = "";
	for (let i = 0; i < quantity; i++) {
		password += characters.charAt(Math.floor(Math.random() * characters.length));
	}
	return password.toLowerCase();
};
```

The alphabet is `[a-z0-9]` after the `toLowerCase()`, so `DATABASE_URL=mysql://louez:${DB_PASSWORD}@mysql:3306/louez` is always well-formed. (`${base64:N}` would be a different story — that one does emit `+` and `/`, which is why this blueprint uses `password`, not `base64`, for anything that ends up inside a URL.)

**2. "Encode config.env in the required array form"**

Both forms are supported. From `packages/server/src/templates/processors.ts`, the template contract types `config.env` as either shape:

```ts
config: {
	isolated?: boolean;
	domains: DomainConfig[];
	env:
		| Record<string, string | boolean | number>
		| (string | Record<string, string | boolean | number>)[];
	mounts?: MountConfig[];
};
```

and `processEnvVars()` has an explicit branch for each (`// Handle array of env vars` / `// Handle object of env vars`). The table form is also the majority convention in this repo today — 304 blueprints use `[config.env]` versus 150 using `env = [...]` — and `CONTRIBUTING.md` documents `[[config.env]]` under the `[config]` section rather than mandating the array.

The stack itself was deployed and exercised end to end before submitting (services healthy, schema installed on first boot, account created, product image uploaded through the app into the MinIO bucket and served back byte-identical), and `node build-scripts/generate-meta.js --check` reports `✅ 515 templates validated`.

Siumauricio:
Tested end-to-end on a real Dokploy instance — everything works on the first try. 🎉

**Deployment (imported via base64 → deploy):**
- `compose.deploy` → `done` in ~50s
- Containers: `louez` (synapsr/louez:2.1.0), `mysql:8.4`, `minio` — all **running**, healthchecks passing (louez only starts after mysql+minio report healthy, and it did)
- Traefik routing: `http://<domain>/` → **200**, real Next.js app (`<title>Louez.io</title>`), `/login` → 200, `/dashboard` → 200. No Traefik 404s, schema installed itself on first boot as advertised.

**Conventions:** no `container_name`, no host ports, no external network, image pinned to `2.1.0`, per-template `meta.json` (id = folder, logo present), secrets via `${password:32}`, domain maps to `louez:3000`. `generate-meta.js --check` → ✅ 515 templates validated.

**Upstream legitimacy:** checked before deploying — [Synapsr/Louez](https://github.com/Synapsr/Louez) is a real, active project: created Jan 2026, ~1000 commits, 9 releases (v1.1.0 → v2.1.0), 8 organic forks spread over months, AGPL-3.0, and `synapsr/louez` on Docker Hub has ~4.8k pulls with a manifest that's publicly pullable and tagged consistently with the releases.

Also agree with the author's rebuttal of the automated review: `${password:N}` only emits `[a-z0-9]` so the `DATABASE_URL` is always well-formed, and the `[config.env]` table form is supported (and is the majority form in this repo).

The MinIO entrypoint note in the PR description is a nice touch — accurate, and useful for future MinIO-based blueprints.

LGTM ✅

PR #1087, opened 2026-08-17
Fix Appwrite Function API requests behind HTTPS redirects

ChiragAgg5k:
## What is this PR about?

Fix the Appwrite template so Functions receive an HTTPS `APPWRITE_FUNCTION_API_ENDPOINT`.

```text
before: POST http://<domain>/v1/... -> 301 -> GET https://<domain>/v1/...
after:  POST https://<domain>/v1/...
```

Dokploy redirects HTTP to HTTPS permanently. Node follows that redirect by changing POST to GET, which can turn a TablesDB `createRow` request into `listRows` because both operations use the same path.

## Checklist

- [ ] I have tested the template in my instance, so the maintainers don't spend time trying to figure out what's wrong.

Validation performed:

```text
python tomllib validation: passed
node build-scripts/generate-meta.js --check: 517 templates validated
Node redirect reproduction: POST followed as GET across the current 301 redirect
```

A Dokploy instance was not available locally. This is a one-line configuration correction; the existing template's HTTPS domain and redirect configuration were verified statically.

Reported in https://appwrite.io/threads/1538184240431370403

## Screenshots or Videos

Not applicable.


Siumauricio:
Tested this on a Dokploy instance (hostinger.dokploy.com) by importing `blueprints/appwrite` from this branch (merged with `canary`). Summary: the change is the canonical one and works behind Dokploy/Traefik **when HTTPS is enabled on the domain**, but it makes the template hard-require an HTTPS domain, so I pushed a small comment to the toml documenting that.

**Why the change is correct**
- Appwrite's own variable docs (`app/config/variables.php`): `_APP_OPTIONS_FORCE_HTTPS` — *"Set to 'enabled' whenever the public API is served over HTTPS, including when TLS is terminated by a reverse proxy."*
- In `appwrite/appwrite:1.9.5` (`app/controllers/general.php` L472-478) `APPWRITE_FUNCTION_API_ENDPOINT` is built as `(_APP_OPTIONS_FORCE_HTTPS == 'disabled' ? 'http' : 'https') + apiHostname`, so this env var is the only lever for the Functions endpoint scheme; `_APP_OPTIONS_ROUTER_FORCE_HTTPS` does not affect it.
- The HTTPS check uses `X-Forwarded-Proto` (utopia-php/swoole `Request::getProtocol()`), which Traefik sets correctly (a spoofed header from the client is overwritten).

**Deploy results** (all 29 containers running, deploy 83 s)

1. Domain with HTTPS enabled in Dokploy (the realistic production setup):
   - `GET https://<domain>/v1/health` → `401 general_unauthorized_scope` JSON (API reached over TLS, no redirect)
   - `POST https://<domain>/v1/account` → `400 general_argument_invalid` JSON (POST accepted over HTTPS, no protocol error)
   - `POST http://<domain>/v1/account` → `308` to https (this is the redirect the PR works around; with the fix the Functions runtime receives an `https://` endpoint and never hits it)
2. Domain **without** HTTPS (what Dokploy creates by default when the template is deployed — `certificateType: none`, `https: false`):
   - `GET http://<domain>/v1/health` → `301 https://<domain>/v1/health` → `404 page not found` (Traefik has no TLS router for the host)
   - `POST http://<domain>/v1/account` → `500 {"type":"general_protocol_unsupported"}`
   - `/console` still serves the static SPA (200) but every API call fails, so login/signup does not work until HTTPS is enabled. On `canary` (`disabled`) the same plain-HTTP deployment works.

So the trade-off is: correct HTTPS behaviour (Functions endpoint, generated URLs) vs. plain-HTTP deployments no longer working out of the box. Given Appwrite's guidance and that HTTPS is the expected setup, I think the change is right; I pushed **92e06a7** to this branch adding a comment next to `_APP_OPTIONS_FORCE_HTTPS` so users know they must enable HTTPS on the main domain (validators still pass, TOML parses). Note: Dokploy's template importer forces `certificateType: "none"` on template domains, so the template itself cannot pre-enable TLS.

Not verified end-to-end: actually deploying a Function and inspecting `APPWRITE_FUNCTION_API_ENDPOINT` inside the runtime (I relied on the 1.9.5 source above).

PR #1089, opened 2026-08-18
Add MyLinks template

Sonny93:
Adds a Dokploy blueprint for [MyLinks](https://github.com/my-links/my-links), a self-hostable bookmark manager (organize, search and share links, collections, click-based ranking, browser extension).

## Template

- `my-links` service (AGPL-3.0, image `sonny93/my-links:5.3.0`) + its own `postgres:18-alpine`, following the same two-service pattern as the existing `linkwarden` template.
- `APP_KEY` and `DB_PASSWORD` generated via Dokploy helpers, `APP_URL` derived from the assigned domain.
- Validated with `node build-scripts/generate-meta.js --check`.

## Testing notes

Not yet deployed through a live Dokploy instance — will test via the PR preview once available and report back before merge.


Siumauricio:
Tested on a live Dokploy instance (hostinger.dokploy.com) from this branch merged with `canary`. Summary: works after one small healthcheck fix, which I pushed to your branch.

**Upstream / legitimacy**
- `my-links/my-links`: org repo created 2022-04, ~986 commits, AGPL-3.0, releases `webapp-v5.3.0` (2026-08-16). PR author @Sonny93 is the sole upstream maintainer (985/986 commits, only org member).
- Image `sonny93/my-links:5.3.0` on Docker Hub (repo registered 2023-11, ~2.6k pulls), manifest -> 200 without auth. Note: the tag is published for `linux/amd64` only; arm64 hosts won't be able to pull it (not a blocker, just FYI).

**Conventions**
- No `container_name` / host `ports` / networks, per-template `meta.json` with logo + links, pinned image, `${base64:32}` for `APP_KEY`, `${password}` for `DB_PASSWORD`, `postgres:18-alpine` with the volume on `/var/lib/postgresql` (correct for pg18). Env set matches upstream `.env.example` / `compose.image.yml`.
- `generate-meta.js --check`, `validate-docker-compose.ts` and `validate-template.ts` all pass.

**Deploy #1 (as submitted): ❌ Traefik returned `404 page not found`**
- Both containers were `running`, migrations ran (`Migrated in 1.3 s`) and the app logged `started HTTP server on 0.0.0.0:3333`, but the `my-links` container stayed **unhealthy** (`FailingStreak 12`), so Traefik dropped it from routing.
- Healthcheck output: `wget: can't connect to remote host: Connection refused`. Cause: on Alpine, `localhost` resolves to `::1` first and busybox `wget` doesn't fall back to IPv4, while the server only listens on `0.0.0.0` (IPv4).

**Fix pushed: 2365e67** – healthcheck now targets `http://127.0.0.1:3333/` instead of `localhost`.

**Deploy #2 (with the fix): ✅**
- `my-links-1` **running (healthy)**, `postgres-1` **running (healthy)**, RestartCount 0 on both.
- Through Traefik on the generated domain: `GET /` -> 200 (`MyLinks`), `/login` -> 200, `/register` -> 200 (email/password form renders).

**Non-blocking nits**
- `version: "3.8"` isn't required by Dokploy (the commit message says it is) — it's harmless, but you can drop it.
- `APP_URL` is `http://${main_domain}`; users enabling HTTPS on the domain should switch it to `https://` (same pattern as the `linkwarden` template, so consistent with the repo).

Ready to merge from my side.

PR #1104, opened 2026-08-23
feat: add postgres-pgadmin template

iamtoricool:
## Add PostgreSQL + pgAdmin template

Adds a blueprint for **PostgreSQL 18 Alpine** with **pgAdmin 4** Web UI, validated with Podman.

### Stack

*   **PostgreSQL** `postgres:18-alpine` — volume `postgres-data:/var/lib/postgresql` (correct for Pg18, not `/data`), healthcheck `pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}`, `expose: 5432`
*   **pgAdmin 4** `dpage/pgadmin4:8.13` — volume `pgadmin-data:/var/lib/pgadmin`, `PGADMIN_DEFAULT_EMAIL/PASSWORD`, `SERVER_MODE=False`, `MASTER_PASSWORD_REQUIRED=False`, `depends_on: postgres healthy`, `expose: 80` via `[[config.domains]]`

### Template

*   `template.toml` variables: `pgadmin_domain=${domain}`, `postgres_user=${username}`, `postgres_password=${password:32}`, `postgres_db=postgres`, `pgadmin_email=${email}`, `pgadmin_password=${password:16}`
*   Single domain for pgAdmin (`serviceName=pgadmin port=80`); Postgres internal via `expose` (add second domain if external TCP exposure desired)
*   `meta.json` `id: postgres-pgadmin` matches folder, `version: 18-alpine` matches postgres tag, logo `postgres-pgadmin.svg`

### Testing

*   `node build-scripts/generate-meta.js --check` → ✅ 518 templates validated
*   Podman (`podman compose up -d`) — `postgres | Up (healthy)` (`PostgreSQL 18.6`, `pg_isready` passing), `pgadmin | Up` (gunicorn `http://[::]:80`), `psql -U testuser -c "SELECT version();"` → `PostgreSQL 18.6`, host `curl 127.0.0.1:5050` (override) → 302 → `/browser/`


Siumauricio:
Tested on a real Dokploy instance (hostinger.dokploy.com), imported from the branch and deployed with fresh volumes. Since `maintainerCanModify` is on, I pushed the fixes below directly to your branch (a29d2fc).

**As submitted**
- pgAdmin domain → `HTTP 200` «pgAdmin 4»; both containers `running`, postgres `healthy`. The non-root pgAdmin volume (`/var/lib/pgadmin`, uid 5050) works fine.
- The second `[[config.domains]]` (`serviceName = "postgres"`, `port = 5432`) → `HTTP 502 Bad Gateway`. Dokploy domains are Traefik HTTP(S) routes; pointing one at the Postgres TCP port can never work (and it silently spawns a second random `${domain}` host). Removed it.
- pgAdmin was not wired to the bundled Postgres at all: users had to add the server by hand, which is the same experience as the existing `pgadmin` template plus Dokploy's native Postgres. A combo template should ship pre-connected.
- `dpage/pgadmin4:8.13` is from Nov 2024; current stable is `9.17`.

**Pushed fix (a29d2fc)**
- Dropped the `postgres` domain/variable.
- Added `[[config.mounts]]` for `servers.json` (host `postgres`, `${postgres_user}`/`${postgres_db}`) and a `pgpass` file with `${postgres_password}`. The pgAdmin service copies the pgpass into the data volume with `0600` before starting (libpq ignores group/world-readable password files, and uid 5050 cannot chmod the root-owned bind mount). Import only happens on first launch (empty `pgadmin-data`), per pgAdmin docs.
- Bumped pgAdmin to `9.17`.

**Re-test after fix**
- pgAdmin → `HTTP 200`; container log: `Added 0 Server Group(s) and 1 Server(s).`
- `GET /browser/server/nodes/1/` → server `PostgreSQL`, host `postgres`, db `postgres`, username = generated `${username}`.
- `POST /browser/server/connect/1/1` → `{"success":1,"info":"Server connected.","data":{"connected":true,"version":180006,"user":{"is_superuser":true}}}` — connects without a password prompt.
- `generate-meta.js --check`, `validate-docker-compose.ts`, `validate-template.ts` all pass.

LGTM after the fix.

Issue #1108, opened 2026-08-22
Odoo template is wrong


BacemJ:
### To Reproduce

Creating a fresh app of odoo (any version) wil not work sometimes without creating any log sometiomes with log like this 

<img width="885" height="720" alt="Image" src="https://github.com/user-attachments/assets/e5b28e3a-0b85-48b3-9494-e48732edc03e" />

### Current vs. Expected behavior

A fresh app should start without issues

### Provide environment information

```bash
dokploy updated
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

Maybe, need help

BacemJ:
Nothing resolved, just deployment error added to previous error of no logs and running container.

<img width="967" height="529" alt="Image" src="https://github.com/user-attachments/assets/4ba74828-c51a-44b7-a85a-f0066473193d" />

<img width="991" height="436" alt="Image" src="https://github.com/user-attachments/assets/eb4308b5-ffa4-46b4-8deb-6e84178fec5b" />

<img width="929" height="443" alt="Image" src="https://github.com/user-attachments/assets/573aeaaa-586b-4eb3-bab9-dba564e939b8" />

<img width="520" height="705" alt="Image" src="https://github.com/user-attachments/assets/e384d517-24a0-4d56-926d-b7a2cf992441" />

BacemJ:
The problem was that the container doesn't have permission to write on the /etc/odoo which prevent Odoo from running the default odoo.conf file which oblige the user to create an odoo.conf from the host vps.
With this volumes configuration the odoo database manager will never work properly.
The second problem is that odoo doesn't have permission to write on the /var/lib/odoo wich another big probleme that prevent odoo from running this probleme is simply resolved by modifying compose file by adding this
```shell
command: >
  bash -c "chown -R odoo:odoo /var/lib/odoo &&
           /entrypoint.sh odoo"
``` 
And when odoo run you will discover problem running POS, and chats and they can be fixed by ading this to the compse file in web server
```
;WebSocket router (8072)
- traefik.http.routers.odoo-pos-ws.rule=Host(`domain.com`) && PathPrefix(`/websocket`)
- traefik.http.routers.odoo-pos-ws.entrypoints=websecure
- traefik.http.routers.odoo-pos-ws.tls.certresolver=letsencrypt
[...]

PR #1110, opened 2026-08-25
data permission right & Traefik configuration

BacemJ:
Permession to write on data directory is mandatory for odoo server to run.
Added Traefik configuration to allow POS and live chat to run

#1108
still problem of write permission on /etc/odoo

## Screenshots or Videos
image

### DIFF
--- a/blueprints/odoo_19/docker-compose.yml
+++ b/blueprints/odoo_19/docker-compose.yml
@@ -11,7 +11,27 @@ services:
       - odoo-web-data:/var/lib/odoo
       - ../files/config:/etc/odoo
       - ../files/addons:/mnt/extra-addons
-
+    command: >
+      bash -c "chown -R odoo:odoo /var/lib/odoo &&
+               /entrypoint.sh odoo"    
+    labels:
+        ;WebSocket router (8072)
+      - traefik.http.routers.odoo-pos-ws.rule=Host(`domain.com`) && PathPrefix(`/websocket`)
+      - traefik.http.routers.odoo-pos-ws.entrypoints=websecure
+      - traefik.http.routers.odoo-pos-ws.tls.certresolver=letsencrypt
+      - traefik.http.routers.odoo-pos-ws.service=odoo-pos-ws
+      - traefik.http.services.odoo-pos-ws.loadbalancer.server.port=8072
+      ;Longpolling router (8072)
+      - traefik.http.routers.odoo-pos-longpoll.rule=Host(`domain.com`) && PathPrefix(`/longpolling`)
+      - traefik.http.routers.odoo-pos-longpoll.entrypoints=websecure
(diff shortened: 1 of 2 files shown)


BacemJ:
This pull request need to be fixed. I added comments with false syntax ; must be #

Siumauricio:
Thanks for digging into this — the odoo_19 template **is** genuinely broken on canary (confirms #1108), but the root cause is different from what this PR addresses, and the PR as written cannot deploy. I pushed a fix to your branch (**f4bb8f7**), tested end to end.

**Why the PR as-written fails**
1. The `labels:` block is invalid compose — `docker compose config` rejects it: `services.web.labels must be a mapping` (the `;WebSocket router (8072)` comment lines turn the block into a scalar+sequence mix). The deploy never starts.
2. Even if valid, manual Traefik labels aren't used in Dokploy templates: routing is generated from `template.toml`, and the hardcoded `Host(\`domain.com\`)`, `websecure` entrypoint and `letsencrypt` certresolver wouldn't match any user's instance.
3. The `chown` isn't the failing piece: the official `odoo:19.0` image runs as uid 100 (`odoo`), so the `chown -R` in `command` executes as non-root; and `/var/lib/odoo` is a named volume that inherits `odoo:odoo` ownership from the image automatically.

**Actual root cause (reproduced locally with the canary compose verbatim)**
Dokploy creates `../files/config` as an empty directory, and bind-mounting it over `/etc/odoo` hides the image's `odoo.conf`. The entrypoint's `wait-for-psql.py` then crashes in a restart loop:
```
configparser.NoSectionError: No section: 'options'
```
That's exactly the "fresh app will not work" from #1108.

**Fix pushed (f4bb8f7)**
- Reverted `docker-compose.yml` to canary (no labels, no chown).
- `template.toml` now ships `config/odoo.conf` via `[[config.mounts]]` (lands in `../files/config`, so `/etc/odoo/odoo.conf` exists again) with `proxy_mode = True` — which covers the Traefik/POS/live-chat intent of this PR: in the image's default threaded mode (workers=0), Odoo serves `/websocket` on 8069 itself, so the single Dokploy domain is enough and no 8072 routing is needed.

**Test evidence (deployed on a clean Dokploy instance)**
- `web` and `db` running, RestartCount 0; logs: `Using configuration file at /etc/odoo/odoo.conf`, `HTTP service (werkzeug) running on :8069`.
- Via Traefik: `GET /` → 303 → `/web/database/selector` → **200** "Odoo" (database creation screen).
- proxy_mode active: access logs show the real client IP (X-Forwarded-For honored).
- `/websocket` requests reach Odoo through Traefik with no extra labels (they log 404 only because no database is initialized yet — bus routes register per database).

Note for maintainers: `odoo_17` and `odoo_18` have the same empty-config-mount bug and deserve the same treatment in a follow-up.

PR #1102, opened 2026-08-22
fix(rustrak): point templates at the official images and current metadata

AbianS:
I maintain Rustrak and submitted these two templates a while back. They have gone stale and are handing users a broken install, so this brings them up to date.

### What was wrong

Both blueprints pointed at `abians7/*` images pinned to `v0.2.1`. The current release is `v0.14.2`, so anyone deploying from Dokploy was getting a build twelve minor versions old from a personal Docker Hub namespace. The metadata linked to the old personal repo, and `rustrak-full` also linked to a docs site that has since moved.

### What changed

| | Before | After |
|---|---|---|
| `rustrak` | `abians7/rustrak-server:v0.2.1` | `rustrak/rustrak-server:latest` |
| `rustrak-full` server | `abians7/rustrak-server:v0.2.1-postgres` | `rustrak/rustrak-server:postgres` |
| `rustrak-full` UI | `abians7/rustrak-ui:v0.1.6` | `rustrak/rustrak-ui:latest` |

- `version` is now `latest`, matching the convention other `:latest` templates use.
- Links point at `github.com/rustrak/rustrak` and `rustrak.github.io/rustrak`.
- Logo replaced with the project's current mark.
- The `rustrak` description carried an estimated `<100MB memory footprint`. It now says 32 MB at idle, measured with `docker stats` on the image this template deploys. The `rustrak-full` description drops the figure entirely, since that stack includes PostgreSQL and I have not measured it.

No changes to `template.toml`, the compose structure, or any other blueprint.

### Tested

Both templates were deployed locally from these exact files before opening this PR.

- `rustrak`: server starts, migrations run, superuser is created, Actix listens on 8080.
- `rustrak-full`: PostgreSQL passes its healthcheck, the server connects and migrates against it, the dashboard starts and reaches the API.
- `docker compose config` valid for both, and `node build-scripts/generate-meta.js --check` passes on all 517 templates.

PR #1115, opened 2026-08-27
feat: add Rustrak SQLite full-stack template

GhaziBenDahmane:
## Summary

- add a full-stack Rustrak template with the web dashboard
- use built-in SQLite storage with a persistent volume
- configure separate dashboard and API domains
- generate the session secret and initial administrator credentials

## Validation

- node build-scripts/generate-meta.js --check
- template structure validator
- Docker Compose structure validator
- docker compose config --quiet


Siumauricio:
Tested this template on a live Dokploy instance — works. ✅

**Duplication check (this was my main question):** canary already has `rustrak` (server-only, SQLite) and `rustrak-full` (server + dashboard + PostgreSQL). This PR is not a duplicate — it fills the one missing combination: **dashboard + built-in SQLite, no external database**. You cannot get that with the existing templates without editing them, so it earns its slot. The compose mirrors `rustrak-full` minus postgres (including the `HOSTNAME=0.0.0.0` fix for the Next.js UI from #1105), which is exactly right.

**Conventions:** No host ports / container_name / networks, `expose` only, secrets and admin credentials generated in the toml, separate UI and API domains pointing at the right services/ports (3000 / 8080). Validators pass (527 templates, compose + template validators).

**Deploy:**
- Both containers up in ~31s.
- UI domain `/` → HTTP 200, "Rustrak - Error Tracking System".
- API domain `/` → HTTP 401 `{"error":"Not authenticated"}` (expected), `/health` → HTTP 200.

**Non-blocking suggestion:** the image is `:latest`; upstream now publishes version tags (`v0.14.9` as of 2026-08-27), so pinning would be nicer — though the existing `rustrak`/`rustrak-full` templates also use `:latest` (kept that way in #1102), so this matches current repo state. LGTM.

PR #1121, opened 2026-08-29
fix(postiz): pass all social provider env vars and fix Temporal healthcheck

Hugo0507:
## What is this PR about?

Fixes for the existing **Postiz** template.

The blueprint documents social provider variables in the Environment tab, but
only Facebook, LinkedIn and X are actually passed to the container. Setting
`INSTAGRAM_APP_ID`, `YOUTUBE_CLIENT_ID` or `TIKTOK_CLIENT_ID` has no effect —
Postiz reads an undefined value and builds the OAuth URL with
`client_id=undefined`, so the provider page returns "this page isn't available".

While deploying this template behind HTTPS I hit a few more issues, all fixed here:

- **Temporal healthcheck never passed.** `tctl` requires an explicit
  `--namespace`, and Temporal does not bind `127.0.0.1`. With it fixed,
  `postiz-app` can wait for `service_healthy` instead of `service_started` —
  it used to crash on boot when Temporal was not ready yet.
- **`BACKEND_INTERNAL_URL` used `localhost`**, which resolves to `::1` while the
  backend only listens on IPv4.
- **`NOT_SECURED` was `true`**, disabling secure cookies on an HTTPS deployment.
- **`NEXT_PUBLIC_FRONTEND_URL` was missing.**
- Added `restart: always` to the Temporal services so they come back after a
  host reboot.

`template.toml` was updated so Dokploy pre-creates the new variables in the
Environment tab.

## Checklist

- [x] I have read the suggestions in the README.md file
- [ ] I have added tests that demonstrate that my correction works or that my new feature works. — _N/A, blueprints have no test harness in this repo._

N/A

## Screenshots or Videos

_(optional)_


MikeSoft (approved):
worked to me

Siumauricio:
Tested end to end on a clean Dokploy instance — the core fixes here are correct and verified. I pushed one amendment to your branch (**5c7edc7**) removing the pieces Postiz doesn't actually read; details below.

**Verified against Postiz source (`gitroomhq/postiz-app`, current main)**
- `INSTAGRAM_APP_ID/SECRET`, `YOUTUBE_CLIENT_ID/SECRET`, `TIKTOK_CLIENT_ID/SECRET` — all real (`instagram.standalone.provider.ts`, `youtube.provider.ts`, `tiktok.provider.ts`) and in upstream `.env.example`. Passing them through compose is the right fix: without it they never reach the container and the OAuth URL gets `client_id=undefined`, exactly as described.
- `BACKEND_INTERNAL_URL` localhost → `127.0.0.1`: correct, this is the well-known IPv6 (`::1`) pitfall.
- Temporal healthcheck: confirmed working (see deploy evidence), and gating `postiz-app` on `service_healthy` is now meaningful.

**Amendment pushed (5c7edc7)**
- `GOOGLE_CLIENT_ID/GOOGLE_CLIENT_SECRET` — removed: Postiz never reads them. "Sign in with Google" uses `YOUTUBE_CLIENT_ID` (`apps/backend/src/services/auth/providers/google.provider.ts`), and Google My Business uses `GOOGLE_GMB_CLIENT_ID` with a `YOUTUBE_CLIENT_ID` fallback. Keeping them in the Environment tab would only mislead users.
- `NEXT_PUBLIC_FRONTEND_URL` — removed: the variable doesn't exist anywhere in the Postiz codebase.
- `NOT_SECURED: "false"` — removed the variable entirely: Postiz only checks truthiness (`!process.env.NOT_SECURED` in `proxy.ts`, `main.ts`, `public.controller.ts`), so the string `"false"` is truthy and would have kept insecure mode ON — a silent no-op of your intended fix. Unsetting it is what actually enables secure cookies.

**Deploy evidence (fresh instance, HTTPS domain)**
- All 6 containers running, RestartCount 0. `temporal` reports **healthy** — the new `tctl --address $(hostname -i):7233 --namespace default workflow list` healthcheck exits 0 (the old `127.0.0.1` one could never pass since auto-setup binds the container IP), and `postiz-app` started only after it, as intended.
- `GET /` → 200 "Postiz Register" via Traefik.
- `POST /api/auth/register` → 200 `{"register":true}` with `set-cookie: auth=<jwt>; ... HttpOnly; Secure; SameSite=None` — secure cookies confirmed on.
- Authenticated `GET /api/user/self` → 200 with the created user.

With 5c7edc7 this is good to merge.

PR #1122, opened 2026-08-30
Add heyo docs template

gpiechnik2:
## What is this PR about?

New PR of [Template Name]

## Checklist


Siumauricio:
Thanks for the contribution! I tested this template on a real Dokploy instance and unfortunately **the deploy fails at build time, 100% reproducibly**, so it can't be merged yet.

## What happens

The template builds from `https://github.com/heyo-sh/heyo-docs.git#main` and passes `SITE_URL=https://<domain>` as a build arg. The Dockerfile scaffolds a fresh project with `bun create @heyo-sh/heyo-docs` (currently `0.1.3`), sets `siteUrl` in `heyo-docs.config.ts`, and runs `bun run build` — which fails during prerender:

```
error: Prerender: Request failed for /_.data: Prerender (data): Received a 500
status code from `entry.server.tsx` while prerendering the `/` path.
error: script "build" exited with code 1
```

## Root cause (verified locally, outside Dokploy)

This is an upstream bug in the `@heyo-sh/heyo-docs@0.1.3` react-router template, triggered by `siteUrl` being set:

- `docker build .` with the **default** `SITE_URL=http://localhost:3000` → the build **succeeds**.
- `docker build --build-arg SITE_URL=https://docs.example.com .` → **fails** with the prerender 500 above.
- Reproduced without Docker/Dokploy at all: `bun create @heyo-sh/heyo-docs docs --template react-router` + uncommenting `siteUrl: "https://docs.example.com"` in `heyo-docs.config.ts` + `bun run build` → same failure.

Since the Dokploy template always injects the user's real domain as `SITE_URL`, **every deployment of this template will fail** until the package is fixed.

## Requested changes

1. Fix the `siteUrl` prerender 500 in `@heyo-sh/heyo-docs` and publish a new release.
2. Pin the build context to a tag or commit instead of `#main` (e.g. `...git#v0.1.4` or a SHA) so template deployments are reproducible and don't break when `main` moves.
3. Ping here once done and I'll re-test.

Everything else looks good: compose/toml/meta pass the repo validators, `expose` instead of host ports, healthcheck on 127.0.0.1, and the domain maps to the right service/port.

PR #1141, opened 2026-09-08
libredb-studio: bump version to 0.16.1 and add healthcheck

yusuf-gundogdu:
## What is this PR about?

Bumps the LibreDB Studio blueprint from the published 0.9.59 to 0.16.1, the current release, and adds the healthcheck @nktnet1 asked for in review.

Three changes:

- `docker-compose.yml`: image tag moves to 0.16.1
- `docker-compose.yml`: adds the healthcheck from the review suggestion, unchanged
- `docker-compose.yml`: adds `AUTH_COOKIE_SECURE=false`
- `meta.json`: version field moves to 0.16.1

On `AUTH_COOKIE_SECURE`: Dokploy generates an HTTP `sslip.io` domain by default. The app marks its auth cookie Secure in production on any non-loopback host, so on a plain HTTP domain the browser drops the cookie and login silently loops. 0.16.1 is the first release that lets the operator override it. Four blueprints already in this repo do the same thing for the same reason, including `inkvoice`, which carries the same comment. Without this line the template installs fine and then fails at first login, and an API level check cannot see it because it is browser behaviour.

Tested before pushing, per the contributing guide. I brought this exact compose up against `ghcr.io/libredb/libredb-studio:0.16.1`:

- container reached `Up (healthy)`, so the healthcheck passes against this version
- `/api/db/health` answered `{"status":"healthy","service":"libredb-studio"}`
- `/api/storage/config` answered `{"provider":"sqlite","serverMode":true}`
- admin login with `ADMIN_PASSWORD` returned `{"success":true,"role":"admin"}`, and a wrong password returned HTTP 401
- the login `set-cookie` header carries `HttpOnly; SameSite=lax` and no `Secure` flag, which is what makes login work on an HTTP domain
- `/app/data` was written on the `libredb-data` volume
- after `docker compose restart` the container returned to healthy and login still worked

Environment variable names, the port and the `/app/data` mount path are unchanged from 0.9.59, so existing deployments keep their data across the upgrade.

This PR was opened on 8 September for 0.14.1. Newer releases shipped while it was open, so the branch is updated in place rather than replaced with a new PR. That is why the branch name still says 0.14.1.

## Checklist

- [ ] I have added tests that demonstrate that my correction works or that my new feature works. Not applicable: there is no test suite for blueprints. The repository validators pass and the deployment test is listed above.

None.

### DIFF
--- a/blueprints/libredb-studio/docker-compose.yml
+++ b/blueprints/libredb-studio/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.8"
 services:
   libredb-studio:
-    image: ghcr.io/libredb/libredb-studio:0.9.59
+    image: ghcr.io/libredb/libredb-studio:0.14.1
     restart: unless-stopped
     environment:
       - ADMIN_EMAIL=admin@libredb.org
--- a/blueprints/libredb-studio/docker-compose.yml
+++ b/blueprints/libredb-studio/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.8"
 services:
   libredb-studio:
-    image: ghcr.io/libredb/libredb-studio:0.14.1
+    image: ghcr.io/libredb/libredb-studio:0.15.0
     restart: unless-stopped
     environment:
       - ADMIN_EMAIL=admin@libredb.org
--- a/blueprints/libredb-studio/meta.json
+++ b/blueprints/libredb-studio/meta.json
@@ -1,7 +1,7 @@
 {
   "id": "libredb-studio",
   "name": "LibreDB Studio",
-  "version": "0.9.59",
+  "version": "0.14.1",
   "description": "The modern, AI-powered open-source SQL IDE. Query PostgreSQL, MySQL, Oracle, SQL Server, SQLite, MongoDB and Redis from your browser.",
   "logo": "libredb-studio.svg",
   "links": {
--- a/blueprints/libredb-studio/meta.json
+++ b/blueprints/libredb-studio/meta.json
@@ -1,8 +1,8 @@
 {
   "id": "libredb-studio",
   "name": "LibreDB Studio",
-  "version": "0.14.1",
-  "description": "The modern, AI-powered open-source SQL IDE. Query PostgreSQL, MySQL, Oracle, SQL Server, SQLite, MongoDB and Redis from your browser.",
+  "version": "0.15.0",
+  "description": "The open-source SQL IDE you host yourself. Query PostgreSQL, MySQL, Oracle, SQL Server, MongoDB, Redis, ClickHouse, Cassandra and more from your browser, 16 engines in one interface.",
   "logo": "libredb-studio.svg",
   "links": {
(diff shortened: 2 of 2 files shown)


blueprints/libredb-studio/docker-compose.yml
nktnet1, line 13: @yusuf-gundogdu could we also add healthchecks?

```suggestion
      - STORAGE_SQLITE_PATH=/app/data/libredb-storage.db
    healthcheck:
      test:
        [
          "CMD",
          "node",
          "-e",
          "fetch('http://127.0.0.1:3000/api/db/health').then(r => process.exit(r.ok ? 0 : 1)).catch(() => process.exit(1))"
        ]
      interval: 30s
      timeout: 5s
      retries: 3
      start_period: 15s
```

nktnet1 (changes requested):

yusuf-gundogdu:
Refreshed to 0.16.0, which is the current release. Nothing else in this PR changed.

yusuf-gundogdu:
One question while this is open, and it is about your review time rather than about this PR.

The image tag here is pinned, so a version bump is a one line change with no judgement in it. Two other stores that carry LibreDB Studio now watch the tag and open that bump themselves, so neither side writes or reviews it any more. One of their maintainers put it plainly: updates are handled automatically.

If you would like the same here, say the word and I will send whatever your setup needs, a renovate entry or the equivalent, as a separate small PR. If you would rather keep it manual, that is fine too and I will keep opening these by hand.

I am asking because this is the kind of change that does not really need a person on either end, and I would rather your attention went to the PRs that do.

blueprints/libredb-studio/docker-compose.yml
yusuf-gundogdu: Added, thank you. I took your suggestion unchanged, and it passes against the current release: the container reaches `Up (healthy)` with `ghcr.io/libredb/libredb-studio:0.16.1`, and `/api/db/health` answers `{"status":"healthy","service":"libredb-studio"}`. The image ships no healthcheck of its own, so the compose level one is the only place it can come from.

While testing I found a second thing and fixed it in the same push. Dokploy hands out an HTTP `sslip.io` domain by default, and the app marks its auth cookie Secure on any non-loopback host in production, so the browser drops the cookie and login loops with no visible error. Adding `AUTH_COOKIE_SECURE=false` fixes it. `inkvoice`, `posthog`, `9router` and `omniroute` already carry the same line for the same reason. The login response now sends `HttpOnly; SameSite=lax` with no `Secure` flag.

The branch is also bumped to 0.16.1 and retitled. Sorry for the delay on the healthcheck.

yusuf-gundogdu:
Updated to 0.16.2, which is the current release.

0.16.2 closes a cross-account authorization defect (GHSA-3wh2-8x78-jfw4, scored 8.8): a connection id supplied in a request could select another session's already-authenticated database provider, so a `user` account could reach a connection restricted to `admin`. It matters on any instance more than one person signs in to.

PR #1143, opened 2026-09-08
feat(elasticsearch): update template to v9 with security

nktnet1:
## What is this PR about?

Update Elastic Search to v9

## Checklist


nktnet1:
@Siumauricio would be nice if `[[config.domains]]` expose an `active` property since Dokploy now supports it.

This is so that templates don't automatically expose certain domains (the elastic-search service in this case - since the frontend is Kibana), but is there so that users can optionally enable if they want to with the correct service and expose port configured).

PR #1166, opened 2026-09-16
feat: add Appwrite 2 template

duducp:
## What is this PR about?

New PR of **Appwrite 2** (`appwrite-2`, Appwrite 2.2.0)

Adds a self-hosted Appwrite 2.2 stack adapted from the official `appwrite/appwrite` 2.2.0 `docker-compose.yml`. The bundled Traefik is replaced with Dokploy domains, so the stack is reachable through the Dokploy reverse proxy instead of Appwrite's own.

> Note: this is a new blueprint. There is an open PR (#1132) that updates the existing `appwrite` blueprint to 2.0.0. Happy to consolidate into that one instead if the maintainers prefer not to keep a separate `appwrite-2` template.

## Services

`appwrite`, `appwrite-console`, `appwrite-realtime`, `appwrite-worker`, `appwrite-task-*`, `postgresql`, `mongodb` (replica set), `redis`, `clickhouse`, `mariadb`, `openruntimes-executor`, `orchestrator`, `appwrite-geo`, `appwrite-autogravity`, `appwrite-assistant`, `appwrite-browser`, plus the optional `separate`/`embedding` profile services.

Appwrite 2.2 uses PostgreSQL as its primary database (and for the vectors DB), MongoDB as the documents DB, and Redis/ClickHouse for cache/usage. `mariadb` is started for parity with upstream but is not used by 2.2.

## Notes

- The compose file as committed previously did not parse: a dangling `networks:` block had turned `gateway`/`appwrite`/`runtimes` into phantom services and duplicated the `appwrite` key. That is fixed here.
- All images are pinned to explicit versions (`appwrite/appwrite:2.2.0`, `appwrite/new:1.1.96-self-hosted`, `appwrite/postgres:0.1.0`, `mongo:8.2.5`, `redis:7.4.7-alpine`, `clickhouse/clickhouse-server:26.4.3-alpine`, `openruntimes/executor:0.29.0`, `ghcr.io/open-runtimes/orchestrator/orchestrator:2.2.0`, etc.).
- `restart: unless-stopped` added to the always-on services and to the `separate`-profile workers.
- `depends_on` conditions added so `appwrite` waits for `postgresql`, `redis` and `appwrite-geo` to be healthy.
- Secrets are generated with the `password:32` helper; the MongoDB keyfile and app user are created at first boot via the mounted `mongo-init.js` / `mongo-entrypoint.sh`.
- **Function/site preview subdomains**: Appwrite serves them on `<sub>.<functionsDomain>` / `<unique>.<sitesDomain>`. Dokploy rejects wildcard hosts in `template.toml`, so `HostRegexp` Traefik labels are added to the `appwrite` service (same approach as the existing `appwrite` blueprint).
- `_APP_ROUTER_AUTO_CERTIFICATES=disabled`, since Dokploy terminates TLS and the bundled Appwrite Traefik is not deployed.
- `_APP_COMPUTE_RUNTIMES_NETWORK=dokploy-network` and the executor's `docker.sock`/`/tmp` mounts follow the same pattern as the existing `appwrite` blueprint.

## Validation

- Deployed from the PR preview on a self-hosted Dokploy instance and verified working.
- `node build-scripts/generate-meta.js --check` passes.
- `docker compose config` passes for the default and for the `separate`/`embedding` profiles.
- Services, images, entrypoints, volumes and env keys were diffed against the official 2.2.0 compose and match; `depends_on` intentionally adds health conditions.

## Checklist

N/A

## Screenshots or Videos

N/A


duducp:
Superseded by #1167 (the head branch was renamed from `feat/appwrite-2-2` to `feat/appwrite-2`, and the blueprint from `appwrite-2-2` to `appwrite-2`).

